PR #3, opened 2019-10-02
Create cta.in2p3.fr.pub

iancollier:
Adding config for CTA repo, per request from DESY

https://ggus.eu/index.php?mode=ticket_info&ticket_id=143407


DrDaveD:
I forgot that deleting the source branch would automatically close the PR, so you can skip that step.

iancollier:
I believe my mistake was to try to do this in the web interface.

PR #5, opened 2019-10-03
Using the CVMFS certs as default for EGI access to LIGO caches

efajardo:
This would help us have it as default that EGI access uses the CVMFS cert bundle. This will help us a lot and if sites want it they can all always override it.

@DrDaveD, @zvada  what are your thoughts?


efajardo:
@DrDaveD I did not know you put this fix in place. So I am happy to close this.

PR #28, opened 2020-07-13
Various updates to the cvmfs-config.cern.ch template

jblomer:


etc/cvmfs/common.conf
```diff
@@ -2,6 +2,6 @@
 # cvmfs does not parse included files for variable names, so first need
 #   to set in the domain config any variable set here.
 
-if [ -n "$CVMFS_HTTP_PROXY" ] && [ "$CVMFS_HTTP_PROXY" != "DIRECT" ] && [ "$CVMFS_HTTP_PROXY" != "auto;DIRECT" ]; then
+if [ -n "$CVMFS_HTTP_PROXY" ] && [ "$CVMFS_HTTP_PROXY" != "DIRECT" ]; then
```
DrDaveD, line 5: This check for "auto;DIRECT" is important, because the WPAD server inserts the backup proxies when they are needed.

etc/cvmfs/default.conf
```diff
@@ -4,3 +4,9 @@ if [ "$(($$ % 2))" == 1 ]; then
 else
   CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat"
 fi
+
+# Automatically detect DNS nameserver changes
+CVMFS_DNS_ROAMING=on
+
+# Remember that the config repository is mounted
+CVMFS_CONFIG_REPO_DEFAULT_ENV=1
```
DrDaveD, line 12: What uses this?

etc/cvmfs/default.conf
```diff
@@ -4,3 +4,9 @@ if [ "$(($$ % 2))" == 1 ]; then
 else
   CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat"
 fi
+
+# Automatically detect DNS nameserver changes
+CVMFS_DNS_ROAMING=on
```
DrDaveD, line 9: Is this something we should be setting for EGI & OSG too?

etc/cvmfs/domain.d/cern.ch.conf
```diff
@@ -1,7 +1,7 @@
 CVMFS_FALLBACK_PROXY=
 . ../common.conf
 
-if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
+if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ]; then
```
DrDaveD, line 4: This change would eliminate the use of Cloudflare whenever WPAD is used, including for the CernVM default and LHC@Home.  Why would you want to do that?  Same question for EGI, OSG, and HSF.

etc/cvmfs/default.conf
jblomer: Maybe.  It is particularly useful for roaming laptops.  It comes with the small overhead of installing a watch point on /etc/resolv.conf.  I guess grid worker nodes would not change the nameserver but the setting should also not hurt.

etc/cvmfs/domain.d/cern.ch.conf
jblomer: Eventually I came to the conclusion that for cvmfs-config-default I don't want to judge based on the proxy settings whether or not to use the OpenHTC servers.  This setting is now explicit (`CVMFS_ENABLE_OPENHTC`) and should be set for the outreach & volunteer use cases.  Our professional users should connect to our normal services.

etc/cvmfs/default.conf
DrDaveD: I made prs #31, #32, & #33 to add this to the other 3 branches.

etc/cvmfs/domain.d/cern.ch.conf
DrDaveD: I strongly disagree but you also feel strongly about it and I wasn't able to persuade you so I'm letting it go.

etc/cvmfs/config.d/atlas-nightlies.cern.ch.conf
```diff
@@ -1,6 +1,5 @@
-if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
+if [ "$CVMFS_USE_OPENHTC" = "yes" ]; then
```
DrDaveD, line 1: Let's change this to CVMFS_USE_CDN as @rptaylor [suggested](https://github.com/cvmfs/cvmfs/pull/2586#issuecomment-658406333), both here, elsewhere in this PR, and [cvmfs PR 2586](https://github.com/cvmfs/cvmfs/pull/2586).

etc/cvmfs/default.conf
DrDaveD: Oh now I see that this already was enabled by default in /etc/cvmfs/default.conf, so why do we want it also in the config repo's default.conf?

etc/cvmfs/common.conf
```diff
@@ -2,6 +2,11 @@
 # cvmfs does not parse included files for variable names, so first need
 #   to set in the domain config any variable set here.
 
-if [ -n "$CVMFS_HTTP_PROXY" ] && [ "$CVMFS_HTTP_PROXY" != "DIRECT" ] && [ "$CVMFS_HTTP_PROXY" != "auto;DIRECT" ]; then
-  CVMFS_FALLBACK_PROXY="http://cvmfsbproxy.cern.ch:3126;http://cvmfsbproxy.fnal.gov:3126"
```
DrDaveD, line 6: It is important to still include the fallback proxies when accessing the known grid domains, because they could be being used from a large site that has its own squids that are failing.  We do not want those sites connecting directly to stratum 1s with lots of worker nodes.

Since you have repurposed common.conf and you may want to use it from other domains sometime, you might want to instead put this in a new config file fallback.conf or something like that.

etc/cvmfs/default.conf
jblomer: Yep, that's unnecessary. Removed.

DrDaveD (approved):
All comments have been addressed.

PR #29, opened 2020-07-15
Update ga tech cache

efajardo:


efajardo:
@zvada can you please review this and apply it?

DrDaveD:
It looks OK to me.  It's unfortunate that they would make us change such a trivial name though, instead of them using the old name as a CNAME DNS alias for the new one.

efajardo:
Hi @DrDaveD ,

This is not just a name change. It is a whole new beefier machine. But I guess we could have done a DNS swap or something.

DrDaveD:
@zvada this is waiting on you

PR #34, opened 2020-08-21
Remove sc-cache.chtc.wisc.edu

matyasselmeci:
The CHTC cache is underpowered and should only be used for local jobs.


matyasselmeci:
@bbockelm said he wanted the cache to stay public so I'm going to close this. I told our user to require `/cvmfs/stash.osgstorage.org` to be mounted in his jobs so downloads will fall back properly in case our cache gets overloaded.

PR #35, opened 2020-09-08
Set CVMFS_FALLBACK_PROXY in case common.conf sets it

DrDaveD:
These settings were left out, and cvmfs currently doesn't notice when CVMFS_FALLBACK_PROXY is set in common.conf.


jblomer (approved):
Many thanks for the careful check!

jblomer:
It's deployed in the `TESTING` tag

PR #36, opened 2020-09-16
Sync osg up with default branch, add CVMFS_CLIENT_PROFILE and CVMFS_USE_CDN

DrDaveD:
This adds support for the CVMFS_CLIENT_PROFILE and CVMFS_USE_CDN settings that were recently added to the default branch (which goes along with the cvmfs-config-default rpm).  Those may be set in /etc/cvmfs/default.local if desired.

The behavior of these new settings are slightly different than the default branch but consistent with the former behavior of the OSG configuration repository.  The differences between the branches are when CVMFS_USE_CDN is not set: if CVMFS_CLIENT_PROFILE is set to single or CVMFS_HTTP_PROXY is set to DIRECT or auto;DIRECT, the osg branch implies using the CDN but the default branch does not.

@jblomer Please review to see if you notice any bugs or any inconsistency with what I wrote as the differences.


etc/cvmfs/default.conf
```diff
@@ -1,9 +1,13 @@
...
+  if [ "$(($$ % 2))" -eq 1 ]; then
+    CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat"
+  else
+    CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat"
+  fi
 fi
 
-# Automatically detect DNS nameserver changes
-CVMFS_DNS_ROAMING=on
+# Clear out default setting from cvmfs-config-osg rpm's defualt.d file so
```
jblomer, line 11: defualt.d --> default.d

etc/cvmfs/default.conf
```diff
@@ -1,9 +1,13 @@
-if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
-    CVMFS_SERVER_URL="http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1osggoc-cvmfs.openhtc.io/cvmfs/@fqrn@"
-    CVMFS_FALLBACK_PROXY=""
-else
-    CVMFS_SERVER_URL="http://cvmfs-s1bnl.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1goc.opensciencegrid.org:8000/cvmfs/@fqrn@"
+if [ -z "$CVMFS_PAC_URLS" ] || [ "$CVMFS_PAC_URLS" = "http://wpad/wpad.dat" ]; then
```
jblomer, line 1: Should we add this guard to the cvmfs-config.cern.ch repository, too?

jblomer (changes requested):
Looks good to me!

(Just a nit found.)

etc/cvmfs/default.conf
DrDaveD: Fixed.

etc/cvmfs/default.conf
DrDaveD: Yes I think so.  Also I think there should be a check there to not set fallback proxies if the user set CVMFS_USE_CDN=yes.  I plan to make a PR after this one for the egi branch, and then another for the default branch to make as much in common between the branches as possible while preserving their intentional differences.

jblomer (approved):

DrDaveD:
I don't want to change the OSG config in this major of a way before leaving on vacation, so I plan to merge this next week.

PR #44, opened 2020-12-10
Update StashCache cache at Wisconsin

djw8605:
Requested by @matyasselmeci


DrDaveD:
Note that the master branch is a superset of the different configuration repositories that nobody uses in production.  Submit another PR against the osg branch to change the osg config repo.

DrDaveD:
This also does need to get into the master branch so I'm reopening.

PR #47, opened 2021-02-15
Add KEK domain configuration

jblomer:

PR #51, opened 2021-03-03
Add KEK domain configuration

jblomer:
Replaces #47


etc/cvmfs/domain.d/kek.jp.conf
```diff
@@ -0,0 +1,13 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1kek-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1ral-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1desy-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1ihep-cvmfs.openhtc.io/cvmfs/@fqrn@"
+else
+    CVMFS_SERVER_URL="http://cvmfs-stratum-one.cc.kek.jp:8000/cvmfs/@fqrn@;http://cernvmfs.gridpp.rl.ac.uk:8000/cvmfs/@fqrn@;http://grid-cvmfs-one.desy.de:8000/cvmfs/@fqrn@;http://cvmfs.racf.bnl.gov:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-stratum-one.ihep.ac.cn:8000/cvmfs/@fqrn@"
```
DrDaveD, line 10: Please use cvmfs-s1fnal.opensciencegrid.org instead of cvmfs.racf.bnl.gov.  The latter is itself an alias to cover an old name for the server.

DrDaveD (changes requested):

DrDaveD (approved):

DrDaveD:
Yes, although it's probably best to deploy it in default and test there a little while first.  osg isn't ready to merge yet but it's ok to have the PR ready and egi should be ready.  I'll also want a PR against the master branch although I usually take care of that myself.

jblomer:
This is now deployed in the /cvmfs/cvmfs-config.cern.ch in the `TESTING` branch. I'd like to also deploy #55 before giving it to the lxplus and lxbatch teams for testing.

olifre:
Many thanks for this effort also from our side — we are amongst the sites asked to mount the KEK repositories (on two resources), but we are not really delighted by the "one-shot-RPM" or "manual configuration" steps that were provided by the experiment. So having something in the config repository which is also updated when Squids change is really welcome :+1:.

PR #52, opened 2021-03-04
Convert osg to accepting worldwide stratum 1s

DrDaveD:
This changes the osg config repo to accept worldwide stratum 1s.  For now it enables both OSG and original domain keys.  This change needs to be deployed before the OSG stratum 1s can switch over to using the original keys instead of re-signing them.


etc/cvmfs/domain.d/desy.de.conf
```diff
@@ -1,10 +1,14 @@
...
+# These are here so cvmfs will notice them if common.conf sets them
 CVMFS_USE_CDN="$CVMFS_USE_CDN"
 CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
 CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
-CVMFS_SERVER_URL="$CVMFS_FALLBACK_PROXY"
-. ../osghosted.conf
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1desy-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1cern-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1ral-cvmfs.openhtc.io/cvmfs/@fqrn@"
```
jblomer, line 8: Just to be sure, are the DESY repositories actually being replicated to CERN and RAL?

etc/cvmfs/domain.d/lsst.eu.conf
```diff
@@ -1,10 +1,17 @@
...
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1cern-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1in2p3lsst-cvmfs.openhtc.io/cvmfs/@fqrn@"
+    CVMFS_SERVER_URL="$CVMFS_SERVER_URL;http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io/cvmfs/@fqrn@"
+else
+    CVMFS_SERVER_URL="http://cvmfs-stratum-one.cern.ch:8000/cvmfs/@fqrn@;http://cclssts1.in2p3.fr/cvmfs/@fqrn@"
+    CVMFS_SERVER_URL="$CVMFS_SERVER_URL;http://cvmfs-s1bnl.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@"
+fi
+CVMFS_PUBLIC_KEY="$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/lsst.eu.pub"
```
jblomer, line 14: Shouldn't we create a domain directory for the key?

etc/cvmfs/domain.d/sdcc.bnl.gov.conf
```diff
@@ -1,10 +1,15 @@
...
-CVMFS_SERVER_URL="$CVMFS_FALLBACK_PROXY"
-. ../osghosted.conf
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1osggoc-cvmfs.openhtc.io/cvmfs/@fqrn@"
+else
+    CVMFS_SERVER_URL="http://cvmfs-s1bnl.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1goc.opensciencegrid.org:8000/cvmfs/@fqrn@"
+fi
+CVMFS_PUBLIC_KEY="$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/sdcc.bnl.gov.pub"
```
jblomer, line 12: Should this key be put in a domain directory?

jblomer (commented):
Looks good to me (but see questions)

jblomer (approved):

etc/cvmfs/domain.d/desy.de.conf
DrDaveD: Yes.  They were already being used from there in egi.

etc/cvmfs/domain.d/lsst.eu.conf
DrDaveD: It wasn't that way in egi, and I think it's best to keep it consistent across branches.  I don't see much value added when there's only a single key.  If a second one is needed later we can always change it.

etc/cvmfs/domain.d/sdcc.bnl.gov.conf
DrDaveD: Likewise to above comment, except it wasn't in the egi branch that way but it was in the master branch (because I put it there recently to work with the new auto key download feature in [manage-replicas](https://github.com/cvmfs-contrib/cvmfs-manage-replicas/blob/master/manage-replicas.conf#L24).  That feature reads either individual keys matching a domain name or a directory matching a domain name.

PR #55, opened 2021-03-10
Add support for unprotected osgstorage repositories in default

jblomer:
Adds the common osgstorage.org configuration to the default branch without the specializations for protected repositories.  In particular, this should give CernVM users access to dune.osgstorage.org

It uses the EGI version of osgstorage.org


DrDaveD:
As it turns out, the EGI configuration for osgstorage.org and for that matter the newly updated OSG configuration are all in need of updating.  RAL, NIKHEF, ASGC, and IHEP are all hosting at least one osgstorage.org repository, and I expect they will be adding more soon.  Please add those stratum 1s to the server urls, copied from opensciencegrid.org.conf.

Also, the master and osg branches now have a /etc/cvmfs/keys/osgstorage.org directory instead of pointing to the opensciencegrid.org repository.  We might as well make default and egi consistent with them as well while we are doing updates.  The opensciencegrid.org.pub in the directory is a symlink to ../opensciencegrid.org/opensciencegrid.org.pub.  I did that in order to better support the new manage-replicas key download feature.

DrDaveD (approved):
Looks good.

PR #57, opened 2021-03-25
Use BEARER_TOKEN_FILE to find token in ligo.osgostorage.org

DrDaveD:
This makes ligo.osgstorage.org token support use the WLCG standard BEARER_TOKEN_FILE for the variable name to use to find the authorizing access token.


djw8605 (approved):
LGTM

also, what an oddly long environment variable name.

PR #68, opened 2021-05-20
Add ligo.hpc.swin.edu.au

djw8605:


djw8605:
Waiting on confirmation from admin that this is the right host.  It's not responding from external IPs (at least from my tests)

DrDaveD:
I think you probably just need to change it port 8443.

djw8605:
Yup, you are right.  But now the cert seems expired.  I'll provide feedback to the site.

djw8605:
@DrDaveD could you review and shepard this change wherever it needs to go?

DrDaveD:
You have confirmed now that it works?

djw8605:
Yup, works.

DrDaveD (approved):

DrDaveD:
Since @zvada is traveling this week I'll go ahead and deploy this.  We should also add this to the egi branch in case people in Australia use the egi configuration, and it should go into master as well.  Each one has to be a separate PR because this part of the config is different in each.  @djw8605 please go ahead and create those PRs as well.

PR #70, opened 2021-05-24
Add ligo.hpc.swin.edu.au

djw8605:


etc/cvmfs/config.d/ligo.osgstorage.org.conf
```diff
@@ -1,5 +1,5 @@
 CVMFS_EXTERNAL_URL="https://stashcache.t2.ucsd.edu:8444//user/ligo/;https://osg-kansas-city-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-chicago-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-new-york-stashcache.nrp.internet2.edu:8444//user/ligo/;https://dtn2-daejeon.kreonet.net:8444//user/ligo/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://fiona.uvalight.net:8444//user/ligo/;https://stashcache.gravity.cf.ac.uk:8444//user/ligo/;https://xcache.cr.cnaf.infn.it:8443//user/ligo/;https://xcachevirgo.pic.es:8443//user/ligo/"
+CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://fiona.uvalight.net:8444//user/ligo/;https://stashcache.gravity.cf.ac.uk:8444//user/ligo/;https://xcache.cr.cnaf.infn.it:8443//user/ligo/;https://xcachevirgo.pic.es:8443//user/ligo/;https://ligo.hpc.swin.edu.au:8443//user/ligo/"
```
DrDaveD, line 2: Actually I'd prefer this one to also be on the previous line because that's the convention I was following, although it wasn't documented.  For that purpose, please add a comment before the first line that says "# subset of stashcache servers configured by OSG" and another comment before the second line that says "# stashcache servers not configured by OSG"

DrDaveD (changes requested):

djw8605:
Updated.

DrDaveD (changes requested):
Please accept my suggested extra comments, so they're all in one PR for Jose to merge and distribute in egi.

DrDaveD (approved):

DrDaveD:
@jose-caballero I will leave this to you to merge and distribute.

jose-caballero (approved):

etc/cvmfs/config.d/ligo.osgstorage.org.conf
```diff
@@ -1,4 +1,6 @@
-CVMFS_EXTERNAL_URL="https://stashcache.t2.ucsd.edu:8444//user/ligo/;https://osg-kansas-city-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-chicago-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-new-york-stashcache.nrp.internet2.edu:8444//user/ligo/;https://dtn2-daejeon.kreonet.net:8444//user/ligo/"
+# subset of stashcache servers configured by OSG
+CVMFS_EXTERNAL_URL="https://stashcache.t2.ucsd.edu:8444//user/ligo/;https://osg-kansas-city-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-chicago-stashcache.nrp.internet2.edu:8444//user/ligo/;https://osg-new-york-stashcache.nrp.internet2.edu:8444//user/ligo/;https://dtn2-daejeon.kreonet.net:8444//user/ligo/;https://ligo.hpc.swin.edu.au:8443//user/ligo/"
```
jblomer, line 2: Just wondering, are the double slashes after the port numbers typos?

DrDaveD: Probably.  I think the first one was set that way and everybody has just been following the same conventions since when they add a new one.

DrDaveD:
But in this case isn't it interpreted by cvmfs instead of xrootd?  I guess it would still turn into an `http GET //...` on the xrootd server.

etc/cvmfs/config.d/ligo.osgstorage.org.conf
DrDaveD: But in this case isn't it interpreted by cvmfs instead of xrootd? I guess it would still turn into an http GET //... on the xrootd server.

PR #72, opened 2021-06-15
Changes for supporting ligo.osgstorage.org in cvmfsexec

DrDaveD:
These changes are part of what's needed to get ligo.osgstorage.org to be mountable under cvmfsexec.  It also needs a current nightly build of cvmfs (or version 2.8.2 when it is released, or greater later) and cvmfsexec version 4.12 (or greater).


zvada (approved):
LGTM

PR #73, opened 2021-08-03
Add colorado.xsede.org.pub key

zvada:
Freshdesk 67319.


DrDaveD:
Rename to remove the "colorado." from the public key, because it's going to apply to the whole domain.  Also add a corresponding domain.d/xsede.org.conf.  I suggest starting by copying domain.d/gridpp.ac.uk.conf but use the three OSG stratum 1s from osgstorage.org.conf: bnl, fnal, and goc.

DrDaveD:
Replaced by #74

PR #77, opened 2021-08-03
Bring egi osgstorage config up to date

DrDaveD:
Corresponds to #49, #53, #71 and #72.


DrDaveD:
@jose-caballero This is low priority, but to do sometime just to keep egi in sync with osg.

jose-caballero:
Hi @DrDaveD, could you squash the last 2 commits?

DrDaveD:
I could, but I don't think it would be an improvement.  In fact I had originally put in one new commit to add that symlink, but I decided to back it out after I discovered these other 2 commits that were already in the repo on another branch that I could cherry-pick.  I think it's better to share commits as much as possible between branches for this special-purpose github repo.

jose-caballero (approved):

PR #80, opened 2021-10-04
Make CVMFS_HIDE_MAGIC_XATTRS=yes the default for sw repos

jblomer:


DrDaveD (approved):
Looks good to  me.  I'm not sure if should be included for osg & egi but we could leave it only in default for a while and see if any problems appear.

DrDaveD:
I assume you will take care of deploying it.

PR #82, opened 2022-01-04
updating chicago cache to use its canonical name instead of its alias

ddavila0:


DrDaveD:
Why update the Chicago name and not the others (kansas-city, new-york, houston, and sunnyvale)?

PR #92, opened 2022-03-22
add server at Edinburgh

jose-caballero:
Add to CVMFS_EXTERNAL_URL the URL for the StashCache server at Edindurgh.

### DIFF
--- a/etc/cvmfs/domain.d/osgstorage.org.conf
+++ b/etc/cvmfs/domain.d/osgstorage.org.conf
@@ -14,7 +14,7 @@ CVMFS_KEYS_DIR=$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/osgstora
 # subset of stashcache servers configured by OSG
 CVMFS_EXTERNAL_URL="http://stashcache.t2.ucsd.edu:8000/;http://osg-kansas-city-stashcache.nrp.internet2.edu:8000/;http://osg-new-york-stashcache.nrp.internet2.edu:8000/;http://dtn2-daejeon.kreonet.net:8000/"
 # stashcache servers not configured by OSG
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://fiona.uvalight.net:8000/;http://stashcache.gravity.cf.ac.uk:8000/;http://xcache.cr.cnaf.infn.it:8000/;http://xcachevirgo.pic.es:8000/;http://stashcache.edi.scotgrid.ac.uk:8000/"
+CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://fiona.uvalight.net:8000/;http://xcache.cr.cnaf.infn.it:8000/;http://xcachevirgo.pic.es:8000/;http://stashcache.edi.scotgrid.ac.uk:8000/"
 CVMFS_EXTERNAL_MAX_SERVERS=4
 CVMFS_FOLLOW_REDIRECTS=yes
 CVMFS_QUOTA_LIMIT=1000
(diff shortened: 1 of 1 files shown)


DrDaveD:
Hold this for a bit.  I have a question in to the Edinburgh site admin if he has any objections if we make the new server the default for all non-LIGO osgstorage.org repositories in the region, and remove Cardiff from the list.

jose-caballero:
This is what I got from the person who requested adding the instance at Edingurgh:

"""
There are some CPU and memory limits at Cardiff, from DUNE's understanding. 

We're assumming now only DUNE and LIGO are using stashcache in the UK. So if Cardiff takes care of LIGO, Edinburgh will take care of the rest (DUNE). 

No idea if we should remove Cardiff from the general config or not.
"""

What about we add Edinburgh for now, in order to move forward, and we remove Cardiff later on if needed?

DrDaveD:
> What about we add Edinburgh for now, in order to move forward, and we remove Cardiff later on if needed?

The problem with leaving Cardiff in is that some DUNE jobs will select Cardiff over Edinburgh if they happen to be closer to Cardiff according to geoIP location.  So please go ahead and remove `http://stashcache.gravity.cf.ac.uk:8000/;` from osgstorage.org.conf.

jose-caballero:
I may be missing something here. 
The comment I got from them included "... Cardiff takes care of LIGO..." If we remove Cardiff from osgstorage.org.conf, how will it deliver files for LIGO to UK sites?

jose-caballero:
Ah, maybe because Cardiff is also listed in ligo.osgstorage.org.conf?

DrDaveD:
> Ah, maybe because Cardiff is also listed in ligo.osgstorage.org.conf?

Exactly.

ddavila0:
I'd like to include the admin of cardiff in this thread, @DrDaveD could you add the github user 'duncanmmacleod' to the repo so that he can comment, please?

brianhlin:
I don't believe this repo locks down who can comment. @duncanmmacleod should be able to weigh in without being added to anything

duncanmmacleod:
> I don't believe this repo locks down who can comment. @duncanmmacleod should be able to weigh in without being added to anything

Mic check, 1 2.

ddavila0:
Ah! my bad, when I type the "@" I didn't see Duncan's username.

DrDaveD:
Ok, so @duncanmmacleod I assume you have no problem with removing the Cardiff stashcache server as a default osgstorage.org cache, leaving it only for ligo.osgstorage.org.  Correct?  This PR will replace the osgstorage.org general service with the stashcache in Edinburgh.

duncanmmacleod:
I'm not sure of the context of the proposed change, and what the implications are.

Is the change independent of the configuration of the stashcache we host in Cardiff, or is there something we can change? (I say we, our instance is entirely managed out of UCSD)

DrDaveD:
It is independent of the stashcache configuration.  This PR is for the cvmfs configuration repository for EGI.  After this is merged, we will copy the change also to the default cvmfs configuration managed by CERN.  It's not relevant to the OSG configuration.

ddavila0:
@duncanmmacleod: the suggested modification would only change the namespace that the cache would be able to serve from. Instead of serving data form all VOs, it would only serve data from LIGO.

This line would have to change from "ANY" to "LIGO" :
https://github.com/opensciencegrid/topology/blob/master/topology/Cardiff%20University/Cardiff%20Computing%20Cluster/CardiffPRPCachingInfrastructure.yaml#L26

jose-caballero:
I have just removed Cardiff from the list of URLs in osgstorage.org.conf
It is still in the dedicated file for LIGO: etc/cvmfs/config.d/ligo.osgstorage.org.con

DrDaveD (approved):

jose-caballero:
Hi Derek @djw8605 
is it OK with you if I issue the merge request? 

Cheers,
Jose

djw8605 (approved):

djw8605:
Yup, I agree with the others.  This is fine with me.

PR #95, opened 2022-04-20
computecanada config for soft repo

### LINKED ISSUE #67: conditions for setting CVMFS_USE_CDN=yes
Hello,

Regarding https://github.com/cvmfs-contrib/config-repo/blob/master/etc/cvmfs/common.conf#L15
As I understand the intention is to ensure a CDN is used, if the use of a CDN had not been previously defined, and no specific proxies are defined.

Before we discussed the CVMFS_USE_CDN variable there was not really a standard approach for this, so we (CC) came up with our own approach to make things simple and convenient for our end users (some of whom may not know much about proxies).  We allowed CVMFS_HTTP_PROXY to remain unset in the client's configuration, and if so, used that as a flag to select use of the CDN (and DIRECT). That way it "just works" out of the box (using CDN) without having to worry about choosing or understanding proxies, while also allowing more advanced users who have proxies available to define and use them (in which case they also get the "real" s1 servers).

I wonder if we could expand this , something like:
```
if [ -z "$CVMFS_USE_CDN" ]; then
  if [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
    # If an explicit site proxy and CDN use/nonuse are not set, use the CDN
    CVMFS_USE_CDN=yes
  fi
  if [ -z "$CVMFS_HTTP_PROXY"  ]; then
    CVMFS_USE_CDN=yes
    CVMFS_HTTP_PROXY=DIRECT
  fi      
fi
```
I think this would more broadly fulfill the same intention. If both CVMFS_HTTP_PROXY and CVMFS_USE_CDN are unset, then use DIRECT and CDN.  It would only affect users with minimal basic config, they would not need to know the details about what CDNs and proxies are and it "just works" for them without having to understand those settings and apply them. It would allow certain of our end users to transition from the CC config repo to the default config repo (another simplification for them allowing it to more easily work out of the box), and no other users would be affected (since everyone else would have already been required to set a valid CVMFS_HTTP_PROXY).

rptaylor:
Add config and key for Compute Canada soft repo.   fix #67 

For now that is the only repo to make externally available. The alternative to this way would be to have a config.d/restricted.computecanada.ca file with a URL like http://null.invalid .  The connections will be blocked but we also want to prevent external clients from initiating connections for that repo.

Is there a testing procedure for updates to the default config repo branch?


etc/cvmfs/domain.d/computecanada.ca.conf
```diff
@@ -0,0 +1,10 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+# See CVMFS_SERVER_URL in ../config.d/soft.computecanada.ca.conf
```
DrDaveD, line 6: Why define domain.d at all if it only works on /cvmfs/soft.computecanada.ca?

etc/cvmfs/keys/computecanada.ca/soft.computecanada.ca.pub
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
DrDaveD, line 1: Is this key really unique to the "soft" repository only?  If not it would be better for it to have the more generic name.  If there are unique keys per repository then it's not a good idea to set CVMFS_KEYS_DIR because then all keys will be accepted for all repositories (in the future, when there is more than one)

etc/cvmfs/domain.d/computecanada.ca.conf
rptaylor: For consistency, to include the common.conf file and set other common config like the keys dir.

etc/cvmfs/keys/computecanada.ca/soft.computecanada.ca.pub
rptaylor: No, this is the shared key used for most of our repos.  It could have a generic name computecanada.ca.pub but some specificity is useful; we do think of it as "the soft key" for repos in "the soft stack". Otherwise we might end up with names like computecanada.ca.pub and computecanada2.ca.pub which are less descriptive.

We generally only use separate keys for repos that are hosted on separate servers.

DrDaveD: Ok but I think it is confusing to have it match the name of the repository.  cern.ch keys has some called `cern-it1.cern.ch.pub` etc, so maybe you could use `soft-repos.computecanada.ca.pub` or `soft-stack.computecanada.ca.pub`.  Or perhaps better yet, make a different subdirectory like `/etc/cvmfs/keys/computecanada.ca/softstack/computecanada.ca.pub` because that could make the directory usable for CVMFS_KEYS_DIR.  It's not good to put keys controlled by different people in the same CVMFS_KEYS_DIR directory since all of the keys will work for all of the repositories.  That's my suggestion.

rptaylor: The keys are all controlled by the same people. The reason for having separate stacks (servers, keys) is just logistical, for example there is a dev one, and a data one with different infrastructure. They are all managed by the same team (us).

>  If there are unique keys per repository then it's not a good idea to set CVMFS_KEYS_DIR because then all keys will be accepted for all repositories

Isn't using multiple keys and multiple repos the only reason for using CVMFS_KEYS_DIR? It sounds like you are saying each keys dir should contain only a single key, which would be indistinguishable from using CVMFS_PUBLIC_KEY (though even that allows defining a list of multiple colon-separated files).  As long as nobody can inject fake keys into the config repo I'm not sure what the risk is.

DrDaveD: I think it's really for the case where multiple keys can sign any/all of the repositories.  That's the way CERN did it, and as far as I know so far they're the only ones using the feature.  Since they're all managed by the same team then it's probably ok to put them in the same directory, since you wouldn't care which key signed which repo.

jblomer (approved):
Apologies for the long delay! Looks good to me!

jblomer:
Regarding testing: yes, I update the cvmfs-config.cern.ch repository on the `TESTING` tag first. So if you add `CVMFS_REPOSITORY_TAG=TESTING` in /etc/cvmfs/config.d/cvmfs-config.cern.ch.local, you get the changes before it goes to production.  We leave it in test usually for 1 week.  CERN pre-production nodes thus have some time to pick up the changes.

I update the cvmfs-config TESTING tag just now with the changes from this PR.

rptaylor:
@jblomer great, thanks!  I was away for a couple weeks.
I set up the TESTING tag and it seems to work as expected.

@DrDaveD perhaps, it could be useful to OSG/EGI users. We provide a very large general purpose comprehensive software stack, so it likely contains numerous scientific software suites that could be of interest to users in a wide variety of disciplines.
https://docs.computecanada.ca/wiki/Available_software
I am not sure what sort of software environment is set up on OSG/EGI clusters or if they are intended for general purpose use; anyway you are welcome to use the soft repo.

BTW I recall you made a computecanada branch of this repo, copied from our config repo, but that is somewhat different from what we wanted to have added here.

rptaylor:
@jblomer can you let me know when it has graduated to the prod tag?  Thanks!

jblomer:
@rptaylor It's available now in prod

PR #96, opened 2022-04-21
Use osg 3.5 wn-client instead of current for LIGO cvmfs-x509-helper libraries

DrDaveD:
As discussed in [SOFTWARE-4799](https://opensciencegrid.atlassian.net/browse/SOFTWARE-4799)

PR #99, opened 2022-04-25
Update libexec interface for LIGO 

DrDaveD:
This cherry-picks #96 which uses the osg3.5 worker node client for missing cvmfs-x509-helper libraries for LIGO, rather than "current", because "current" will soon be pointing to osg3.6 which does not have some of the libraries (that is, the GSI libraries).

It also cherry-picks #97 which eliminates a wrapper process on cvmfs_scitoken_helper by calling exec.


jose-caballero:
Note: I am on holidays. I will have a look next week, once I am back to work.

DrDaveD:
Hopefully this can be done soon, as I believe osg3.5 is now end of life which means the `/cvmfs/oasis.opensciencegrid.org/mis/osg-wn-client/current` symlink could be updated to 3.6 anytime.

jose-caballero:
today we celebrate Labor Day in Europe.

jose-caballero (approved):
OK

PR #102, opened 2022-06-07
Add config and keys for domain name cta-observatory.org

jose-caballero:


etc/cvmfs/domain.d/cta-observatory.org.conf
```diff
@@ -0,0 +1,2 @@
+CVMFS_SERVER_URL="http://cvmfs-stratum-one.zeuthen.desy.de:8000/cvmfs/@fqrn@;http://cccvmfsctas1.in2p3.fr/cvmfs/@fqrn@"
+CVMFS_KEYS_DIR=/etc/cvmfs/keys/cta-observatory.org
```
jblomer, line 2:
```suggestion
CVMFS_KEYS_DIR=$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/cta-observatory.org
```

DrDaveD:
I began to compose a response to this yesterday but apparently never submitted it and now can't find it.  Yes there's Jakob's suggestion, and also a missing inclusion of ../common.conf and associated variable settings.  See skatalescope.conf for an example.

In addition, I see a problem with the cvmfs-stratum-one.zeuthen.desy.de stratum 1 because it appears to be round-robin between two independent stratum 1s, which doesn't work correctly.  Who is the administrator?  Fortunately it is already included in the frontier-squid list of major stratum 1s, due to a [ggus ticket from 2019](cvmfs-stratum-one.zeuthen.desy.de).

jose-caballero:
I have implemented the suggested changes. 

The request came from andreas.haupt@desy.de

DrDaveD:
Let's wait since the stratum 1 host is going to move to the other stratum 1 at DESY.

jose-caballero:
I am going to close this pull request, and open a new one from a dedicated branch "add_cta_observatory". This is to prevent having my local branch "egi" locked until this work is done.

PR #103, opened 2022-07-05
Add iris

jose-caballero:
Add config file and public key for a new domain name .iris.ac.uk
As there isn't yet any actual repository, starting with only the RAL's Stratum-1 is fine.


DrDaveD (changes requested):
Looks like cta-observatory to me.  Wrong files.

jose-caballero:
yes, something didn't work as I expected.

PR #104, opened 2022-07-11
Add configuration for new domain name iris.ac.uk

jose-caballero:
Add the public key and config file for brand new domain name iris.ac.uk
As this is only for the UK, and there is currently no repository with that domain name, it is fine for the time being to use only the Stratum-1 at RAL.


DrDaveD (changes requested):

etc/cvmfs/domain.d/iris.ac.uk.conf
```diff
@@ -0,0 +1,3 @@
+. ../common.conf
```
DrDaveD, line 1: Please add the 4 lines of prefix before including common.conf that is in all the other domain.d/*.conf files.

DrDaveD (changes requested):

jose-caballero:
Fixed.

DrDaveD (approved):

jose-caballero:
Should I issue the Merge myself, or someone else from the CVMFS core team have to do it?

DrDaveD:
You do it, now that it is approved.

PR #105, opened 2022-07-12
Add geoapi and send info header settings where missing

DrDaveD:
These appear to have been inadvertently left out of two domains each.   We're getting lots of requests to the UCSD stashcache from osgstorage.org repositories at CERN because geoapi is disabled on osgstorage.org, so please distribute soon.


jblomer (approved):

DrDaveD:
@jblomer Please merge and distribute

jblomer:
It's in `TESTING`, since it is a small change I'll deploy tomorrow to production

PR #115, opened 2022-09-02
invert computecanada URL logic to support more repos beyond soft

rptaylor:
As contemplated in https://github.com/cvmfs-contrib/config-repo/pull/95, we anticipate having more repos of public interest.

So instead of only specifying soft, we set the URLs for all repos in the domain.d file, but block the restricted repo explicitly with a config.d file.

Please let me know when this is available with the testing tag  CVMFS_REPOSITORY_TAG=TESTING @jblomer .


jblomer (approved):

jblomer:
@rptaylor It's now available in `TESTING`

rptaylor:
Thanks @jblomer , I checked with CVMFS_REPOSITORY_TAG=TESTING  and it still works for /cvmfs/soft.computecanada.ca/

jblomer:
Thanks for confirming! It's now released in production.

PR #116, opened 2022-09-16
Add GSI base configuration (e.g., for fairsoft.gsi.de)

jblomer:
Some comments
- Sören should approve the PR
- The stratum 1 is not listening on port 8000, that should probably be resolved
- The fairsoft.gsi.de root catalog should probably be further split


DrDaveD:
How about replicating this to at least one other WLCG stratum 1?

gsi-sfleisch:
> How about replicating this to at least one other WLCG stratum 1?

At the moment that is not needed.

gsi-sfleisch:
> The stratum 1 is not listening on port 8000, that should probably be resolved

The external Stratum 1 now listens on port 8000 as well as 80.


> The fairsoft.gsi.de root catalog should probably be further split

OK, I informed the repository owner.

DrDaveD:
> >  How about replicating this to at least one other WLCG stratum 1?

> At the moment that is not needed.

So, it's still in testing, not production use?  Then are there a lot of client sites that will be testing?  If there's only a few maybe for now the clients should just have locally installed configuration.

DrDaveD (changes requested):
Please update the PR with port 8000.

Note that this is going in to the master branch, which is not actually installed in any configuration repo.  Did you intend to do it against the default branch?

jblomer:
I thought we would always target the default branch first and from there move it to the others. I can of course rebase to the default branch.

DrDaveD:
(I believe by "default branch" you mean the "master" branch.  Unfortunately for discussion it's confusing that the "default" branch is not the "default branch".)

Usually in this repo we put things first in the specific distribution's branch, and then from there we (usually I) put it also in the master branch as a superset of what's in the specific branches.  It doesn't work the way normal software development projects work.

gsi-sfleisch:
> How about replicating this to at least one other WLCG stratum 1?

Would this be a "free" operation or does it imply GSI replicating some WLCG repositories?

DrDaveD:
> > How about replicating this to at least one other WLCG stratum 1?
> 
> Would this be a "free" operation or does it imply GSI replicating some WLCG repositories?

It would be free to you, nothing that GSI needs to do.  It just involves one or more of the WLCG stratum 1s replicating your repository.  You would need to make your stratum 0 open to the internet if it isn't yet, so the stratum 1s can read it.  You can put in .htaccess files to restrict the readers if you want to (many other stratum 0s do).

jblomer:
Replaced by #120

PR #119, opened 2022-10-13
Update ligo.osgstorage.org.conf - osg-sunnyvale-stashcache.nrp.internet2.edu

biozit:
Unifying the OSDF/PRP cache ports


biozit:
this is working :-)

DrDaveD (approved):

DrDaveD:
Ok, so this also needs to be changed with a separate PR (perhaps cherry-pick) against the osg branch.  Please go ahead and submit that PR next.

The master branch is intended to be a superset of the 3 configuration branches osg, egi, and default.  That's the branch we use for the audit with OSG's topology.  However, the 3 branches are the ones that get synced with the configuration repositories used in production by osg, egi, and the CVMFS default config.  We don't list all stashcaches in all 3 repos, and this one is not in egi or default.  In fact, default does not support ligo.osgstorage.org at all because it is such a complex configuration.

PR #121, opened 2022-10-14
increase CVMFS_NFILES for alice.cern.ch to 256k

jblomer:


DrDaveD:
CMS has also run into this issue quite a bit.  I wonder if we should do the same for more than just alice.

jblomer:
We can of course add cms.cern.ch.  But I guess you are more thinking of adding it to the cern.ch domain config?
I think the new default is save but I'd still like to see it on real sites for a limited number of repos -- we are coming closer to the kernel limit, which often is at 1M or 2M.

jblomer:
This is now released in `TESTING`

PR #124, opened 2022-10-14
Update ligo.osgstorage.org.conf - default port

biozit:
caches are using ports 8444 and 8443, setting the default port (8443)


DrDaveD (changes requested):
No, we discussed putting in both ports for a transition period, and increasing the number of CVMFS_EXTERNAL_MAX_SERVERS during that period so the number of real servers that a cvmfs client will get won't be decreased below 4.

biozit:
Hi Dave,
   
    I changed the config on caches, now they can provide auth files using ports 8444 and 8443. If we changed the config, the cache still work for both ports 8444 and 8443. After the config is applied, I will remove port 8444 from the PODs.

DrDaveD (approved):
Oh!  That's a better way to do it than what I was thinking.  I didn't know that was an option. 

I'll need you to also submit corresponding changes to the egi and master branches.

DrDaveD:
Oh I see you already did master in #123, so we only need one for egi.

biozit:
Hi Dave, I created the other PR.

DrDaveD:
@jthiltges This is waiting on you to merge and distribute as well.

PR #125, opened 2022-10-14

biozit:
caches are using ports 8444 and 8443, setting the default port (8443)
I changed the config on caches, now they can provide auth files using ports 8444 and 8443. If we changed the config, the cache still work for both ports 8444 and 8443. After the config is applied, I will remove port 8444 from the PODs.


DrDaveD (approved):
This is the egi version of #123 and #124.  It will be up to @jose-caballero to merge.

PR #128, opened 2022-10-17
Update ligo.osgstorage.org.conf - unifying the OSG topology names with CVMFS caches

biozit:


DrDaveD (approved):
This is the same as #126 but for the egi branch.  I will go ahead and merge it, but there's no need to distribute it right away because it is just changing aliases for the machine name, and the audit is against the master branch.

PR #130, opened 2022-10-20

biozit:
Is there any reason for the osg-sunnyvale-stashcache.t2.ucsd.edu if not listed here?


DrDaveD:
No, we don't need to add it to the egi configuration.

PR #134, opened 2022-10-25
updating Sunnyvale hostname

biozit:


jthiltges (approved):
LGTM. Will get it applied to production tonight or tomorrow.

DrDaveD:
This is an import of #133.

There's no rush to publish it.  It's just changing one alias for another for the same machine.

jthiltges:
Applied to production for config-osg.

PR #135, opened 2022-10-25
Update osgstorage.org.conf

biozit:


DrDaveD (approved):
This is an import of #133.

There's no rush to distribute it since it's just changing one alias of a host into another alias for the same host, so I will merge it myself.

DrDaveD:
Wait a minute.  I assumed this was the egi branch.  Where did add-CO-xsede come from?  It must have been leftover from long ago.  I will delete it.  Sunnyvale isn't in the egi branch so nothing futher needs to be done.

PR #137, opened 2022-11-02
fixing Fiona cache hostname

biozit:


DrDaveD (approved):
This is just a change of from one alias to another of the same machine, in order to pass an audit compared to what is registered in OSG topology.  I'll leave it to @jblomer to merge, but note that there's no need to distribute this update since there's no operational change.

jblomer:
Thanks! I'll merge and distribute together with the next change.

PR #141, opened 2022-12-19
Update osgstorage.org.conf

biozit:


etc/cvmfs/config.d/ligo.osgstorage.org.conf
```diff
@@ -1,4 +1,4 @@
...
+CVMFS_EXTERNAL_URL="http://scsd-cache.nationalresearchplatform.org:8000;http://unl-cache.nationalresearchplatform.org:8443;http://mghpcc-cache.nationalresearchplatform.org:8443;https://xrootd-local.unl.edu:1094//user/ligo/;https://stashcache.t2.ucsd.edu:8443//user/ligo/;https://its-condor-xrootd1.syr.edu:8443//user/ligo/;https://osg-kansas-city-stashcache.nrp.internet2.edu:8443//user/ligo/;https://osg-chicago-stashcache.nrp.internet2.edu:8443//user/ligo/;https://osg-new-york-stashcache.nrp.internet2.edu:8443//user/ligo/;https://dtn2-daejeon.kreonet.net:8443//user/ligo/;https://osg-houston-stashcache.nrp.internet2.edu:8443//user/ligo/;https://osg-sunnyvale-stashcache.t2.ucsd.edu:8443//user/ligo/;https://ligo.hpc.swin.edu.au:8443//user/ligo/"
```
DrDaveD, line 1: This is supposed to be https and port 8443, right?  For that matter, I see that the last two added at the front here are also set to http, but I believe they should be https.

DrDaveD (changes requested):

etc/cvmfs/config.d/ligo.osgstorage.org.conf
DrDaveD: Oh and it is spelled wrong, scsd instead of sdsc.

DrDaveD: Should this replace the cache at ucsd?  Likewise for the unl-cache added previously, should it replace the other unl.edu cache?

biozit: sorry, we have 3 new caches:
 - sdsc-cache.nationalresearchplatform.org
 -  unl-cache.nationalresearchplatform.org
 - mghpcc-cache.nationalresearchplatform.org

biozit: Should I make these changes on the others branches?

DrDaveD: Once it is merged here, please make another PR for the master branch with the same changes. 

I don't think we put these changes into egi or default branches, although I see there is stashcache.t2.ucsd.edu in those branches.  Would we want to replace that with sdsc-cache.nationalresearchplatform.org?  In general we expect those two configuration branches to be used in Europe, but we leave a few in those branches in the U.S. just in case somebody is running them there.

You didn't directly answer my other question.  Should these new caches replace the ones at ucsd and unl in the osg config?  Why would we want to list two at the same universities? (The sdsc is at ucsd, for those who don't know.)  

Oh, I see the old one at unl.edu is only for ligo, and it is actually LIGO's origin server, not a stashcache.  There has already been discussion about removing that from the cvmfs list, but I don't know if a decision was reached.

biozit: > Should this replace the cache at ucsd? Likewise for the unl-cache added previously, should it replace the other unl.edu cache?

No, we are keeping those caches and adding these new ones.

biozit: > Once it is merged here, please make another PR for the master branch with the same changes.
> 
> I don't think we put these changes into egi or default branches, although I see there is stashcache.t2.ucsd.edu in those branches. Would we want to replace that with sdsc-cache.nationalresearchplatform.org? In general we expect those two configuration branches to be used in Europe, but we leave a few OSDF servers in the U.S. in those branches just in case somebody is using those configurations in the U.S. or nearby.
> 
> You didn't directly answer my other question. Should these new caches replace the ones at ucsd and unl in the osg config? Why would we want to list two at the same universities? (The sdsc is at ucsd, for those who don't know.)
> 
> Oh, I see the old one at unl.edu is only for ligo.osgstorage.org, and it is actually LIGO's origin server, not a stashcache. There has already been discussion about removing that from the cvmfs list, but I don't know if a decision was reached.

The sdsc is on UCSD as well. However, right now we no going to remove any cache, is that ok for you?

DrDaveD: I guess it's OK, if it's a conscious choice.

DrDaveD (approved):
I'll leave this to @jthiltges to merge and deploy.

Since there are 5 commits to change two little files, I recommend either the PR originator (Fabio) to collapse them down to one commit and do a force-push, or for the merger (John) to do a "Squash and Merge" type of merge, using the pull-down menu by the merge button.

etc/cvmfs/config.d/ligo.osgstorage.org.conf
biozit: ok, thank you

biozit:
@jthiltges Could you do the "Squash and Merge"?

jthiltges:
Testing of changes looked fine. Applied to production. Apologies for the winter break delay.

PR #143, opened 2023-01-20
Add dtn-pas nrp OSDF servers (osg)

biozit:


DrDaveD:
Where are these two servers actually located?  Their IP addresses are very close.  If I put the addresses into the Maxmind GeoIP demo web page, they both show up in Kansas, which is the geographical center of the U.S. (the default U.S. location if Maxmind doesn't know any more details).  We should get the Maxmind Geolite2 City DB to recognize their real locations and give it about 4 weeks to propagate to the stratum 1s before putting them into service.

In fact I went back and tried all the existing registered servers and found these additional ones listed in Kansas:
```
sdsc-cache.nationalresearchplatform.org
mghpcc-cache.nationalresearchplatform.org
```
We need to get those updated ASAP.  They get submitted via the [maxmind form for correcting a geolocation](https://www.maxmind.com/en/geoip-location-correction).  They only want country, city, and postal code, nothing more accurate than that.  Can you do that?

biozit:
Hi,

sdsc-cache.nationalresearchplatform.org - San Diego,
CAmghpcc-cache.nationalresearchplatform.org - Holyoke, MA


  I sent the correction using the form.


*--Fábio Andrijauskas*

DrDaveD:
What about the two new servers?  Where are they?  Did you send in corrections for the two new servers in addition to the two old ones?

biozit:
dtn-pas.cinc.nrp.internet2.edu
Internet2 c/o Lumen
400 Pike Street
Cincinnati, OH 45202
317-278-6622

dtn-pas.bois.net.internet2.edu.
"Internet2 c/o Lumen
1103 Miller St
Boise, ID 83702
317-278-6622"

I was wrong,  I sent the correction using the form.

biozit:
sdsc-cache.nationalresearchplatform.org - localization  ok 
mghpcc-cache.nationalresearchplatform.org - localization ok 
dtn-pas.cinc.net.internet2.edu -  localization ok 
dtn-pas.bois.net.internet2.edu -  localization ok

biozit:
@DrDaveD Should I remove this PR and create new one?

DrDaveD:
Something definitely needs to be done about it, because the base is all off and there's a ton of conflicts.  Either recreate and do force-push, or close and create a new one.

PR #147, opened 2023-01-24
update Korea cache - osg

biozit:


DrDaveD (approved):
The new name is an alias for the old name.

@jthiltges I'll leave it to you to merge and distribute.

DrDaveD:
@biozit Please rebase on the current osg branch.

DrDaveD (changes requested):
I would rather not have all those extra blank lines.

Also, it's better if each of these small changes is a single commit, with a description of what it's doing (that is, essentially the same as the PR title, although usually without the branch name).  I usually back out commits with `git reset HEAD^1`, rebase the current branch, commit the change, and do `git push -f` to force-push.

PR #154, opened 2023-02-07
Set cache offline and update FQDN - OSG branch

biozit:
Removing due to maintenance: xcache.cr.cnaf.infn.it and stashcache.gravity.cf.ac.uk

Updating the Korea cache name: daejeon-kreonet-net.nationalresearchplatform.org


jthiltges:
Changes deployed.

PR #157, opened 2023-02-15
Adding new cache - osg-stash-sfu-computecanada-ca.nationalresearchplatform.org - egi branch 

biozit:
We have a new Canada cache: osg-stash-sfu-computecanada-ca.nationalresearchplatform.org

### DIFF
--- a/etc/cvmfs/config.d/ligo.osgstorage.org.conf
+++ b/etc/cvmfs/config.d/ligo.osgstorage.org.conf
@@ -1,7 +1,7 @@
 # subset of stashcache servers configured by OSG
 CVMFS_EXTERNAL_URL="https://stashcache.t2.ucsd.edu:8443//user/ligo/;https://osg-kansas-city-stashcache.nrp.internet2.edu:8443//user/ligo/;https://osg-chicago-stashcache.nrp.internet2.edu:8443//user/ligo/;https://osg-new-york-stashcache.nrp.internet2.edu:8443//user/ligo/;https://daejeon-kreonet-net.nationalresearchplatform.org:8443//user/ligo/;https://ligo.hpc.swin.edu.au:8443//user/ligo/"
 # stashcache servers not configured by OSG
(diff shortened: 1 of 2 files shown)


DrDaveD:
Oh and this should instead be added to the next line where it says "stashcache servers not configured by OSG".

biozit:
> Oh and this should instead be added to the next line where it says "stashcache servers not configured by OSG".

why?

DrDaveD:
> > Oh and this should instead be added to the next line where it says "stashcache servers not configured by OSG".
> 
> why?

Oh I guess the comment is a little misleading.  It means that it's not configured in the osg branch.

biozit:
done

DrDaveD (changes requested):
The changes look good now, but please squash them down to one commit that says "Add osg-stash-sfu-computecanada-ca" or some variant of that, describing what's changed.

biozit:
done

DrDaveD (changes requested):
It is in one commit now, but the commit message is a mess.  Please simplify it.

biozit:
done

DrDaveD (changes requested):
I'm sorry to be such a stickler, but it's still not accurate.  The first line is too generic, and the second line only applies to ligo.  It's important for the first line to be descriptive.  Extra lines are optional if they add more explanation.  So just replace both lines with something like `Add osg-stash-sfu-computecanada-ca`.

jose-caballero:
https://cbea.ms/git-commit/

biozit:
No problem, please. Could you check if it is ok?

*--Fábio Andrijauskas*

DrDaveD (changes requested):
Unfortunately it is still not ok, because now the first line has `Adding new cache - osg branch` which is not only incorrect since it is the egi branch, it is too generic.  The second line is good; just use the second line as the first line.  You don't need to mention the branch name in the commit, because that's obvious from where it is.  It's good to have the branch name in the PR title, as you did.

biozit:
np, done.

*--Fábio Andrijauskas*

DrDaveD (approved):

DrDaveD:
Ok for @jose-caballero to merge & deploy.

@biozit Also please cherry-pick this to the master branch.

jose-caballero (approved):

PR #159, opened 2023-02-22
Add back UK cache - OSG Branch

biozit:
Add the UK cache cf-ac-uk-cache.nationalresearchplatform.org


DrDaveD (changes requested):
Please turn this into one commit.

Is this really adding it back?  If not, maybe we don't want this in the OSG list. We keep only a small subset of caches outside of North America in the OSG list. If it is adding back, then in egi & default this should have been on the first line of caches that OSG uses, rather than on the line of extra caches that are not used by OSG.

biozit:
nevermind

PR #166, opened 2023-05-18
Add virgo storage osg

djw8605:

PR #168, opened 2023-06-16
Add virgo storage egi

DrDaveD:
This imports #166 to the egi branch


AndresTanasijczuk:
May I ask... What is the criteria to chose the OSDF caches that appear in the CVMFS_EXTERNAL_URL list? Why the CNAF cache (ds-914.cr.cnaf.infn.it) is not included?

djw8605:
@AndresTanasijczuk the CNAF cache is in OSG's topology (and appears correctly configured there), and therefore it should be able to cache ligo / igwn / virgo files.  I think it's just a matter that we registered in topology but forgot to add it over here to CVMFS.

DrDaveD:
> @AndresTanasijczuk the CNAF cache is in OSG's topology (and appears correctly configured there), and therefore it should be able to cache ligo / igwn / virgo files. I think it's just a matter that we registered in topology but forgot to add it over here to CVMFS.

I would suggest adding that in a separate PR from this one, in order to have a good record of adding it and to make for easy cherry-picking to the master branch where we keep a superset of the branch configurations.

jose-caballero (approved):

PR #175, opened 2023-08-15
Add sampa stratum 1 (default)

DrDaveD:
This adds to the default branch the newly fully populated SAMPA stratum 1 as a source for CERN, EGI, HSF, IGWN (via symlinks to OSG config), KEK, and OSG.  It is a cherry-pick of #174 except leaving out DESY since it was not included in the default branch.


DrDaveD:
@jblomer Please review.  I decided to start with the `default` branch before `osg` and `egi` so it would get an independent review.

PR #190, opened 2023-09-11
Adding back stashcache.ligo.caltech.edu

biozit:
Adding back stashcache.ligo.caltech.edu


DrDaveD:
Is this because the geoIP has been corrected?  We should wait a month after it is fixed in the Maxmind database, to leave enough time for all the stratum 1s to be updated.

Also, please don't add more unnecessary blank lines.

DrDaveD:
I just checked and the FNAL stratum 1 still seems to locate this somewhere in Europe so we need to wait a few more weeks until all the stratum 1s are updated.

DrDaveD:
For the record, the algorithm for downloading the DB on stratum 1s with default settings is exactly:

- If it has been 14 or more days and less than 28 days since the last update, try to download once on Tuesdays (in the 10:00 hour local time)
 - If it has been 28 or more days, try once a day.

So if there are no download errors, all the stratum 1s should be updated within 3 weeks of a correction.

DrDaveD:
This is included in #197, closing.

PR #196, opened 2023-09-26
Optimize open file descriptor settings for ALICE

jblomer:


etc/cvmfs/config.d/alice.cern.ch.conf
```diff
@@ -0,0 +1,9 @@
+# The ALICE software framework uses many processes (not only multi-threading).
+# Each processes mapped libraries trigger open files to the cvmfs cache.
+# On many-core nodes, this can overflow the 128k default limit for cvmfs
+# processes.
+# See https://github.com/cvmfs/cvmfs/issues/3067 for details.
+CVMFS_NFILES=524288
```
DrDaveD, line 6: I suggest checking to make sure you're not decreasing the amount, only increasing.

etc/cvmfs/config.d/alice.cern.ch.conf
```diff
@@ -0,0 +1,9 @@
...
+# As of cvmfs 2.11, the refcounted cache manager keeps the number of open 
+# file descriptors of the cvmfs2 process under control.
+CVMFS_CACHE_REFCOUNT=true
```
DrDaveD, line 9: Why set this for only alice.cern.ch?  Is it only temporarily here for wider testing and you intend to later set it everywhere?

DrDaveD (changes requested):

etc/cvmfs/config.d/alice.cern.ch.conf
jblomer: Good point, fixed.

etc/cvmfs/config.d/alice.cern.ch.conf
jblomer: Yes, it is meant to be the future default. Just carefully rolling it out. I clarified in the comment.

DrDaveD (approved):

Issue #185, opened 2023-08-31
OSDF cache inconsistencies between branches


DrDaveD:
The master branch is expected to have a superset of the other branches.  The domain.d/osgstorage.org.conf and osgstorage-auth.conf files in the master branch have sections for servers in osg and servers not in osg.  So I expect to be able to do a git diff on those files between branches, for example from the master branch
```
git diff osg etc/cvmfs/domain.d/osgstorage.org.conf
```
and then the list of external servers in the osg section should be identical between the two branches.  That is not the case, there are a lot of differences.  Likewise for egi and the section of servers not used by osg.  Please make them consistent.  I expect there to be some missing servers in the egi branch from the osg section and for the egi section to be completely missing in the osg branch.  The default branch should be just like egi in osgstorage.org.conf.

@biozit

biozit:
sure!

DrDaveD:
Fixed in #224

PR #195, opened 2023-09-22
Fixing Master

biozit:
Fixing Master

Issue #192, opened 2023-09-11
Inconsistencies between osg and egi branches in lists of OSDF servers


DrDaveD:
In the egi branch I should be able to do `git diff osg domain.d/osgstorage.org.conf` and `git diff osg osgstorage-auth.conf` and see only difference lines beginning with `-` in the North America section of `subset of authenticated OSDF caches used by OSG`, and all the lines after `extra authenticated OSDF caches not used by OSG` should begin with `+`.   There are a few cases where that is not so.  Please make it consistent.

In addition, `git diff default domain.d/osgstorage.org.conf` should show no difference of servers between egi and default branches.

DrDaveD:
In addition, doing `git diff master domain.d/osgstorage.org.conf` and `git diff master osgstorage-auth.conf` should show no differences in the lines after `extra authenticated OSDF caches not used by OSG`. 

Currently this is not the case, because `cf-ac-uk-cache.nationalresearchplatform.org` is in the master branch but not in the egi branch for both files.

Beyond that, the comparison between osg and egi now looks good, but the default branch's `domain.d/osgstorage.org.conf` still has a lot of differences that need to be brought in sync.

DrDaveD:
Also, from the osg branch doing `git diff master domain.d/osgstorage.org.conf` and `git diff master osgstorage-auth.conf` should show no differences in the section `OSDF caches used by OSG` and the lines after `extra OSDF caches not used by OSG` should be marked with `-` only.

Currently there are two servers in the OSG section in the master branch that are not in the osg branch in both files: `mghpcc-cache.nationalresearchplatform.org` and `sdsc-cache.nationalresearchplatform.org`.  sdsc-cache was removed temporarily in #229.  Perhaps it is time to add it back?  If not, it should get removed from the master branch.  Or, instead of deleting it, probably it would be better to comment it out with a dated note that it is temporary.  Likewise mghpcc-cache and #235.

DrDaveD:
Now that #241 is finished, please sync up the default branch with the egi branch

DrDaveD:
Oh and `cf-ac-uk-cache.nationalresearchplatform.org` is still in the master branch's domain.d/osgstorage.org.conf but not in egi.  Shouldn't it be there?   It is in osgstorage-auth.conf in both branches.

DrDaveD:
> Now that #241 is finished, please sync up the default branch with the egi branch

[...]

PR #197, opened 2023-09-26
Adding new config set to OSG

biozit:
Adding the new config set to OSG

### DIFF
--- a/etc/cvmfs/domain.d/osgstorage.org.conf
+++ b/etc/cvmfs/domain.d/osgstorage.org.conf
@@ -14,28 +14,40 @@ else
 fi
 CVMFS_KEYS_DIR=$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/osgstorage.org
 
-
-CVMFS_EXTERNAL_URL="http://unl-cache.nationalresearchplatform.org:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://stashcache.t2.ucsd.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://its-condor-xrootd1.syr.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-kansas-city-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-chicago-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-new-york-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://stash-cache.osg.chtc.io:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-gftp2.pace.gatech.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://daejeon-kreonet-net.nationalresearchplatform.org:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-houston-stashcache.nrp.internet2.edu:8000/"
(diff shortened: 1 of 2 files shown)


etc/cvmfs/domain.d/osgstorage.org.conf
```diff
@@ -14,28 +14,41 @@ else
...
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://its-condor-xrootd1.syr.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-kansas-city-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-chicago-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-new-york-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://stash-cache.osg.chtc.io:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-gftp2.pace.gatech.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://daejeon-kreonet-net.nationalresearchplatform.org:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-houston-stashcache.nrp.internet2.edu:8000/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;http://osg-sunnyvale-stashcache.t2.ucsd.edu:8000/"
+# OSDF caches used by OSG
```
DrDaveD, line 17: We don't need this comment line in the osg branch, because by definition everything here is used by OSG.

etc/cvmfs/osgstorage-auth.conf
```diff
@@ -9,26 +9,46 @@ X509_CERT_DIR=
 
 CVMFS_MAX_RETRIES=0
 
-CVMFS_EXTERNAL_URL="https://unl-cache.nationalresearchplatform.org:8443/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://xrootd-local.unl.edu:1094/"
-CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://stashcache.t2.ucsd.edu:8443/"
+# authenticated OSDF caches used by OSG
```
DrDaveD, line 12: Same here, delete this comment line for this branch.

DrDaveD (changes requested):

etc/cvmfs/domain.d/osgstorage.org.conf
biozit: done

etc/cvmfs/osgstorage-auth.conf
biozit: done

biozit:
@DrDaveD, @jthiltges  do you think we can start the deployment today?

DrDaveD (approved):

jthiltges:
From https://github.com/cvmfs-contrib/config-repo/pull/195#issuecomment-1735807087
> > I am considering adding all the European caches to OSG. Is there any side effect for it?
> 
> Yes, the entire list has to be sent to the stratum 1s for looking up every longitude and latitude, and it's not very likely that the osg configuration will be used in Europe so it seems wasteful. Also potentially it will be a little less maintenance effort in that when new European caches are added they won't necessarily needed to be added to the osg branch.

Are we going ahead with the Europe and other non-Americas caches in the osg branch?

biozit:
> From [#195 (comment)](https://github.com/cvmfs-contrib/config-repo/pull/195#issuecomment-1735807087)
> 
> > > I am considering adding all the European caches to OSG. Is there any side effect for it?
> > 
> > 
> > Yes, the entire list has to be sent to the stratum 1s for looking up every longitude and latitude, and it's not very likely that the osg configuration will be used in Europe so it seems wasteful. Also potentially it will be a little less maintenance effort in that when new European caches are added they won't necessarily needed to be added to the osg branch.
> 
> Are we going ahead with the Europe and other non-Americas caches in the osg branch?

We have new caches in London and Amsterdam, I would like to add more chances to use them. All the caches have been on the right GeoIP location for 3 weeks.

jthiltges (approved):
Thanks for the info Fabio. LGTM. Will merge and distribute shortly.

jthiltges:
Changes published to config-osg as revision 121.

DrDaveD:
This was the first stage of fixes #185 and #192.

PR #201, opened 2023-10-10
Temporarily remove asgc stratum 1

DrDaveD:
The ASGC stratum 1 is hopelessly behind in updates right now, so remove it until it catches up.


jblomer (approved):

jblomer:
Should this be deployed urgently? Otherwise I'd put in in `TESTING` now, deploy it on Thursday.

DrDaveD:
The stratum 1 has been languishing for quite a while, but it is in bad shape.  I would just go ahead and deploy it.

PR #202, opened 2023-10-11
Temporarily remove asgc stratum 1 (osg)

DrDaveD:
This is a cherry-pick of #201 to the osg branch.


jthiltges:
Published to config-osg as revision 122.

PR #206, opened 2023-11-13
add configuration for eessi.io

boegel:
After a lot of preparatory work, we now finally have a new CernVM-FS repository in place for the [European Environment for Scientific Software Installations (EESSI)](https://eessi.io/docs), which is part of the effort to make EESSI ready for production that is funded via the [MultiXscale EuroHPC Centre-of-Excellence](https://www.multixscale.eu).

The contents of this repository will gradually be expanded over the coming weeks/months/years via the semi-automated contribution workflow we have implemented and [documented](https://eessi.io/docs/adding_software/opening_pr/).
Later this year we would like to announce this repository as being the production-ready setup of EESSI.

It would be great to have this configuration be part of the default CernVM-FS configuration repository, especially since we will be using EESSI as a running example in the upcoming [*"Best Practices for CernVM-FS in HPC"* tutorial](https://event.ugent.be/registration/cvmfshpc202312).

cc @bedroge, @ocaisa, @casparvl, @trz42


DrDaveD (changes requested):
I suggest changing the key name to eessi.io.pub, so it implies that it applies to all repositories of the eessi.io domain.  I know there's only one so far but there may be more in the future.

boegel:
@DrDaveD If that's not required, I think we prefer to keep it as is for now, to leave the door open to use a different key per repository (unless that doesn't make much sense).

I guess we can easily change `software.eessi.io.pub` to `eessi.io.pub` later, or would that be a hassle?

Getting the configuration for `software.eessi.io` repository is important for the upcoming tutorial, so we would prefer if that can be done quickly.

etc/cvmfs/domain.d/eessi.io.conf
```diff
@@ -0,0 +1,8 @@
+# Stratum 1 servers for the eessi.io domain
+CVMFS_SERVER_URL="http://aws-eu-central-s1.eessi.science/cvmfs/@fqrn@;http://azure-us-east-s1.eessi.science/cvmfs/@fqrn@"
+
+# Public keys for the eessi.io domain
+CVMFS_KEYS_DIR="$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/eessi.io"
```
bedroge, line 5: @boegel As mentioned on Slack, I think I do agree with @DrDaveD's suggestion: since we're using a Yubikey now, we will be using the same key for all `*.eessi.io` repos, so it makes sense to reflect that in the filename of the public key. We should probably even move it a directory up, as there's no need for having a separate directory anymore for just a single key? So:


```suggestion
# Public key for the eessi.io domain
CVMFS_PUBLIC_KEY="$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/eessi.io.pub"
```

DrDaveD: I would have no objection to moving it up a level, but all the other domains in the default config repo have a directory so I think Jakob at least would prefer it in a directory.  Let's stick with existing convention.

boegel: Indeed, see also `etc/cvmfs/keys/opensciencegrid.org/opensciencegrid.org.pub`, so I'll stick to `CVMFS_KEYS_DIR` and only rename `etc/cvmfs/keys/eessi.io/software.eessi.io.pub` to `etc/cvmfs/keys/eessi.io/eessi.io.pub`

bedroge: Oh, I actually saw an example that did use the `keys` directory (https://github.com/cvmfs-contrib/config-repo/blob/master/etc/cvmfs/domain.d/skatelescope.eu.conf, https://github.com/cvmfs-contrib/config-repo/tree/master/etc/cvmfs/keys), but leaving it in a subdirectory is okay with me as well :-)

DrDaveD (approved):
This looks good to me.  It's up to @jblomer now to review, merge, and deploy.

After he has approved I'll also want a cherry-pick PR to the master branch.  For egi and osg it probably requires a request from somebody in one of the supported VOs in those projects.

jblomer (approved):

PR #209, opened 2023-11-22
Update for the new current osg-wn-client from el7 to el8 for portable files

DrDaveD:
such as the CA certificates.


jthiltges:
Deployed to config-osg as revision 124.

PR #212, opened 2023-12-15
reduce warnings/failures for sft-nightlies repo

rptaylor:
Based on https://cvmfs-monitor-frontend.web.cern.ch/sft-nightlies.cern.ch  and the atlas-nightlies.cern.ch.conf file.

According to my tests it avoids all the "failed to access" warnings because this repo is not hosted on all servers.


DrDaveD:
This looks OK to me but I suggest you submit it first to the default branch to see if Jakob agrees, and get it merged there.  The master branch is a superset of all the branches, so I would put it here if it's also going into default.

rptaylor:
Okay now?

PR #216, opened 2024-01-11
Change nikhef.nl stratum 1 DNS alias

DrDaveD:
From klei to cvmfs01.  Import of #214.


jthiltges:
Published to config-osg as revision 125.

PR #229, opened 2024-02-19
Removing SDSC - network issues

biozit:
Removing SDSC - network issues


jthiltges (approved):
LGTM. I'll hold off a bit for an approval from Dave ahead of the merge and deploy.

jthiltges:
Changes deployed as revision 127

PR #241, opened 2024-03-27
change hostname to match the certificate, and sync up with master branch (egi)

biozit:
change hostname to match the certificate


DrDaveD (changes requested):
Ok, so you are also trying to do part of #192 here.  Good catch on noticing the missing uvalight.net server in osgstorage-auth.conf, I didn't notice that.  Now however I see it is also missing from osgstorage.org.conf, so please add it there too.

biozit:
done

DrDaveD (changes requested):
Oh, but do a diff against the osg or master branch, please.  The server is in a different order in osgstorage.org.conf, and in order to be able to know what's different we want them to have the same order.

biozit:
done

DrDaveD (approved):
Ok, now it is up to @jose-caballero to merge and deploy

jose-caballero (approved):

PR #242, opened 2024-04-01
Added ardc.edu.au domain

hzhu16:


DrDaveD (changes requested):
Thanks for your contribution!

First, note that getting this into the master branch doesn't actually get it used anywhere.  There are 3 major configuration repository distributions, the `default` one that comes with the cvmfs project, `osg` for the Open Science Grid based in the U.S., and `egi` for the European Grid Infrastructure based in Europe.  Once accepted into master it will need to be cherry-picked to the other desired distributions.  I think osg at least will require it to be requested by one of their registered VOs, and am not sure about the others.

Next, is this expected to be used outside of the ardc.edu.au University?  If not, it would probably be better for it to just be in the local cvmfs configurations instead of in a widely-distributed configuration repository.  How about outside of Australia?  If it will be shared primarily in Australia, it should probably be hosted on second stratum 1 in Australia as a backup.  Swinburne would be the logical place, because they host a big stratum 1 with replicas of all the big CVMFS domains.  If it will be used outside of Australia, it should probably get replicated to more stratum 1s in other continents.

etc/cvmfs/domain.d/ardc.edu.au.conf
```diff
@@ -0,0 +1,13 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+# Stratum 1 servers for the ardc.edu.au domain
+CVMFS_SERVER_URL="http://203.101.226.164/cvmfs/@fqrn@"
+
+# Public keys for the eessi.io domain
```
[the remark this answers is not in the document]
hzhu16: Thanks, fixed the comment.

djw8605:
The goal for this repo is 2 things:
1. Replicate this stratum 1 to the OSG stratum 1 servers.  I hope that is accomplished in this file: https://github.com/opensciencegrid/topology/blob/master/virtual-organizations/Neurodesk.yaml
2. Add the repo to the config-repo so it's more widely accessible.

This goal is to increase the accessibility of this repo for users outside of Australia.

DrDaveD:
I see.  At this point do we only care about getting it into the osg distribution for use in North America, or is there a need for it worldwide?

At minimum, the OSG stratum 1s should be added to the list of CVMFS_SERVER_URLs.   For that you could use the example at https://github.com/cvmfs-contrib/config-repo/blob/osg/etc/cvmfs/domain.d/sdcc.bnl.gov.conf#L7.  Include the setting of CVMFS_USE_CDN at the top of the file.   Oh, it's a problem for CDN mode that the original server is an IP address only and not a DNS alias that can be kept constant.  I suggest that we instead include at least the swinburne stratum 1 and ask the administrator of that stratum 1 to add this domain, once it is replicated to the OSG stratum 1.  You can find the names for that stratum 1 for example in https://github.com/cvmfs-contrib/config-repo/blob/osg/etc/cvmfs/domain.d/opensciencegrid.org.conf.

djw8605:
I'll ask if they have a DNS entry for it.  Odd that they only gave us an IP.

DrDaveD:
A reverse DNS lookup gets a name with the IP address encoded in it so it's not much better.

stebo85:
Dear all, thank you for the feedback.

I now added a DNS name for our Stratum 0 server: http://stratum0.neurodesk.cloud.edu.au/

Thank you
Steffen

DrDaveD:
Ok then I added an openhtc.io (Cloudflare CDN) alias using that DNS name.  The alias I used was `s1ardc-cvmfs.openhtc.io`.  So @hzhu16 you can use `http://s1ardc-cvmfs.openhtc.io/cvmfs/@fqrn@` with CDN and `http://stratum0.neurodesk.cloud.edu.au/cvmfs/@fqrn@` without CDN for the CVMFS_SERVER_URL.

Can anyone say if this is to be replicated to only OSG stratum 1s, or do we want it to go further?  Where will jobs be run that use it?

hzhu16:
Thank you all for providing information and instructions. The config file was updated and please let me know if it is right.

etc/cvmfs/domain.d/ardc.edu.au.conf
```diff
@@ -0,0 +1,17 @@
...
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+# Stratum 1 servers for the ardc.edu.au domain
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL=" http://s1ardc-cvmfs.openhtc.io/cvmfs/@fqrn@"
+else
+    CVMFS_SERVER_URL="hhttp://stratum0.neurodesk.cloud.edu.au/cvmfs/@fqrn@"
```
DrDaveD, line 11: typo: change `hhttp` to `http`.

etc/cvmfs/domain.d/ardc.edu.au.conf
```diff
@@ -0,0 +1,17 @@
...
+
```
DrDaveD, line 12: Missing `fi` statement

etc/cvmfs/domain.d/ardc.edu.au.conf
```diff
@@ -0,0 +1,17 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+# Stratum 1 servers for the ardc.edu.au domain
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL=" http://s1ardc-cvmfs.openhtc.io/cvmfs/@fqrn@"
```
DrDaveD, line 9: There should not be an extra blank before the `http`.

DrDaveD (changes requested):
Since this is going to be distributed worldwide, we're going to want to use all the same CVMFS_SERVER_URL settings as are in etc/cvmfs/domain.d/opensciencegrid.org.conf.

That list includes a big stratum 1 in Australia.  Once we have it replicated there I'm unsure about whether or not it makes sense to leave the ardc.edu.au stratum0 in the list, but let's go ahead and at least leave it there for now.  I recommend copying the CVMFS_SERVER_URL lines from opensciencegrid.org.conf and adding a line on the end in each section to append these to the list, the way that we currently append the s1goc stratum 1.  That way it's easier for us to do a 'diff' between other files and see the real differences.  The order is always selected by GEO API anyway so operationally it makes no difference what order they are in the list; we can choose the order that is most convenient for maintenance.

djw8605:
Hi @stebo85, is this repo already replicated to some stratum 1 somewhere already?

stebo85:
Yes, our neurodesk stratum 0 repository is replicated to 3 stratum 1 servers in Australia:
cvmfs-brisbane.neurodesk.org
cvmfs-perth.neurodesk.org
cvmfs-sydney.neurodesk.org

We currently also run Stratum 1 servers in Europe and in the US, but we would like to stop hosting them ourselves and use the OSG servers instead to reduce our costs. 

Thank you so much for everyone's help with this!!!

DrDaveD:
Oh then we definitely shouldn't list your stratum 0 in CVMFS_SERVER_URL.  Would you like us to list your 3 Australian stratum 1s, or do you hope to decommission some of those as well?

stebo85:
Can you list our "primary" Stratum 1: cvmfs-brisbane.neurodesk.org - this one will be guaranteed to be available for the next 5 years. 

Thank you!

hzhu16:
Hi Dave, please review the changes. Thank you!

DrDaveD (approved):

DrDaveD:
Please make a new PR to cherry-pick this to the osg branch

PR #248, opened 2024-04-16
Adding kagra-dsr-b1.icrr.u-tokyo.ac.jp cache

biozit:
Adding kagra-dsr-b1.icrr.u-tokyo.ac.jp cache


DrDaveD:
Are we sure that this cache is only supposed to be used for authenticated, and not also for unauthenticated?

DrDaveD (approved):

PR #253, opened 2024-04-30
Setting Denver cache to use pelican/osdf cache

biozit:
Setting Denver cache to use pelican/osdf cache


matyasselmeci (approved):

DrDaveD:
I am confused.  Why use https for unauthenticated data?

biozit:
Using the Pelican Software, the ports are the same for both services.

matyasselmeci:
It looks like we can use both https and plain http on that port if that's better.

DrDaveD:
That doesn't help me.  Why change the Denver cache to use https port 8443 for CVMFS?

Separately, you do have me curious too: why does Pelican use the same port for both services?  Are you saying it always uses https on port 8443 for authenticated and unauthenticated services, or that it always uses port 8443 for both http and https?

matyasselmeci:
Pelican only has a single instance of xrootd, listening on port 8443. This can be used for both auth and unauth access; if you don't need to pass a credential, you can use plain http on that same port.

DrDaveD:
Oh, this is Pelican server software.   I didn't know that was a thing, I only heard about the client.

So then the question is why should the server software be non-standard?  Why not have it also listen on port 8000, for http?  If nothing else it will save a lot of effort in changing the cvmfs configuration over one by one as the servers are converted.  As it is there will be transition problems as the servers are cut over, because cvmfs only reads the configuration at mount time.  A repository can stay mounted for weeks or even months.

DrDaveD:
It's also possible to set up iptables to forward one port to another.  I assume nftables has a way too although I haven't yet tried it.

biozit:
Let me try something using k8s tricks


*--Fábio Andrijauskas*

biozit:
I can't use k8s tricks here, we are checking the best solution.

bbockelm:
@biozit - can you open a second Kubernetes port in the deploy and map it to the same port inside the container?  That'll provide the backward compatibility that @DrDaveD is looking for.

In the meantime, I think it'd be good to merge this to try and discover any issues.  If we map both ports to the single port inside the container, we could do a single subsequent update for all the other caches.

biozit:
Brain, all OSDF k8s hosts are using “hostnetwork,” we can't do that


*--Fábio Andrijauskas*

DrDaveD:
> In the meantime, I think it'd be good to merge this to try and discover any issues. 

If we need to merge it temporarily I'd rather have it be http than https.  Also this is on the master branch and it needs to be on the osg branch in order to be ready to deploy.  If it's just a temporary change then we might as well make it only on the osg branch.

> If we map both ports to the single port inside the container, we could do a single subsequent update for all the other caches.

One change is better than lots, but if it is set up to listen on both 8000 and 8443, wouldn't we just switch this one back and leave it on 8000?  It's very unusual/surprising to use http on port 8443.

biozit:
@bbockelm @DrDaveD, could we approve this PR to start the tests?

DrDaveD:
I would rather have a PR against the osg branch that uses http instead of https.

bbockelm:
We can do that - but it's going to guarantee that there's going to be subsequent updates.  Long-term, we plan to transition to HTTPS-only.

DrDaveD:
>  Long-term, we plan to transition to HTTPS-only.

What's the point of doing that for CVMFS, which is already guaranteeing the integrity at a higher level?

biozit:
https://github.com/cvmfs-contrib/config-repo/pull/254

PR #254, opened 2024-05-03
Changing port from Denver cache #toPelican

biozit:
Changing the port to use the Pelican cache.


DrDaveD (approved):

DrDaveD:
I leave it to @jthiltges to merge and deploy

DrDaveD:
I think the Pelican software should instead be changed to be compatible, to support http on 8000 and https on 8443.

PR #257, opened 2024-05-13
Add sampa stratum1 to atlas-nightlies and sft-nightlies (default)

DrDaveD:
The sampa stratum1 is also hosting {atlas|sft}-nightlies.cern.ch so we might as well add them.


DrDaveD:
@vvolkl or @jblomer this is still waiting on you

vvolkl (approved):

PR #258, opened 2024-05-13
Add sft-nightlies config and add sampa s1 to atlas-nightlies (osg)

DrDaveD:
This brings the cern.ch nightlies configs in sync with the master branch as of #256.  It avoids needing to search through all stratum 1s for sft-nightlies.cern.ch which is only hosted on 3 stratum 1s.  It may help with some pilot timeouts that ATLAS has been experiencing when checking for that repository.


DrDaveD:
@jthiltges this is up to you to approve, merge and deploy.

PR #259, opened 2024-05-13
Add sft-nightlies config and add sampa s1 to atlas-nightlies (egi)

DrDaveD:
This adds a config for the sft nightlies limiting it to the stratum 1s that it is hosted on, to reduce search time when mounting, bringing into sync with the master branch as of #256.


DrDaveD:
@jose-caballero this is for you to approve, merge, and deploy.

PR #269, opened 2024-09-12
Add compute canada to OSG config

PerilousApricot:
@hellermf and I would like to use compute canada's EasyBuild distro at Vanderbilt. @djw8605 here's the PR we talked about on slack. (I believe these are the changes needed, can't think of how to test this...)


etc/cvmfs/domain.d/computecanada.ca.conf
```diff
@@ -0,0 +1,15 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ] || [ "$CVMFS_HTTP_PROXY" = "DIRECT" ] || [ "$CVMFS_HTTP_PROXY" = "auto;DIRECT" ]; then
+  CVMFS_SERVER_URL="http://cvmfs-s1.computecanada.net/cvmfs/@fqrn@"
+  CVMFS_USE_GEOAPI=no
+else
+  CVMFS_SERVER_URL="http://cvmfs-s1-arbutus.computecanada.ca:8000/cvmfs/@fqrn@;http://cvmfs-s1-beluga.computecanada.ca:8000/cvmfs/@fqrn@;http://cvmfs-s1-graham.computecanada.ca:8000/cvmfs/@fqrn@;http://cvmfs-s1.computecanada.net/cvmfs/@fqrn@"
```
DrDaveD, line 10: I see this was copied from the `default` config-repo, but I think it may be a mistake to include the cloudflare alias cvmfs-s1.computecanada.net in this setup because it will probably confuse the Geo API, and depending on the luck of where you query it from, it may get sorted to be before some of the direct stratum 1s.

@jblomer @rptaylor do you have any comment?  Maybe I'm wrong, maybe the cloudflare alias always ends up last.  I just tried some tests and that appears to be the case.  If so, I think it is accidental.

rptaylor: cvmfs-s1.computecanada.net  is a load balancer of all the other stratum 1 servers listed here (and one other also). It uses [dynamic steering](https://developers.cloudflare.com/load-balancing/understand-basics/traffic-steering/steering-policies/dynamic-steering/) to determine the closest origin. If an origin server is unavailable, the [zero-downtime](https://blog.cloudflare.com/new-tools-to-monitor-your-server-and-avoid-downtime/) feature dynamically redirects a client to another origin within the same HTTP request to avoid serving an error to clients.
 
 I don't think there is a great solution to reconciling the different ways that CVMFS and Cloudflare do geo-location, but the impetus for including it in CVMFS_SERVER_URL (in the case when proxies are used) is mainly for DDoS resilience. When I did some performance microbenchmarks comparing the Cloudflare stratum server to other servers, sometimes it could be faster, sometimes slower, depending on the client location and access patterns. Overall I think the case for including it outweighs the possibility that the connection might (or actually might not) be slightly sub-optimal, and anyway for use outside Canada I think the potential upside should be higher.

DrDaveD: My main motivation to avoid going through Cloudflare when using a local proxy is that there is no logging for debugging problems and watching for (usually accidental) abuse.  Especially if the local proxy goes down and all the traffic goes directly to the stratum 1, I want to be alerted of that as soon as possible to make sure it gets repaired.  Do you have that issue?  We additionally have monitored fallback proxies and configure clients to use those and never to go directly to servers when there is a local proxy.

We have so far not seen any intentional denial of service attack on stratum 1s.

rptaylor: In this case OSG sites would not have access to the logs anyway (whether for stratum servers or Cloudflare), since they are operated by CC.  Shielding against abuse (whether usually accidental or otherwise) and load spikes from proxy failures is another reason to include Cloudflare, isn't it?

We do have a log analytics system (filebeat and Kibana I believe) that could be used to look for anomalous connections but in practice it hasn't been necessary and we haven't had any of those kinds of operational issues.

The other main impetus for using Cloudflare was to extend our reach and presence to anywhere in the world the repo is used, whereas our stratum servers are only located in Canada. It sounds like operating an OSG stratum server that replicates the CC repo(s) would provide what you are looking for (regional network connectivity and access to log files for debugging), which is fine if you want to do that, although using Cloudflare seems simpler and easier to me.

DrDaveD: There might be good reasons for replicating the repos further around the world, but I wasn't thinking about us needing the logs from your servers, I was thinking that you might have good reason to need them.  We experience site proxy failures every week in the WLCG and are interested in not only protecting our servers but also protecting the WAN usage of our sites (and also partly wanting to avoid causing enough hits onto Cloudflare that they will require us to move up from the free tier).   We get this already by our monitoring of the fallback proxies and our stratum 1s.

In any case it looks to me like having the Cloudflare entry here is unlikely to have much effect since it is always geosorted to the end, so it's fine.

DrDaveD (approved):

rptaylor:
@PerilousApricot Can you please include this file?  https://github.com/cvmfs-contrib/config-repo/blob/default/etc/cvmfs/config.d/restricted.computecanada.ca.conf
We want that to prevent external clients from attempting to access the restricted repository (there are module files in the soft repo which reference the restricted repository, which can not easily be separated).

PerilousApricot:
@rptaylor done!

rptaylor (approved):

DrDaveD:
@jthiltges Please merge and deploy

jthiltges:
Changes published to config-osg as revision 141.

PerilousApricot:
@rptaylor sorry for messaging you here, but I don't have other contact info. We (vanderbilt) are interested in hosting a stratum-1 for the repo -- is there a good list for these discussions? I only saw the -announce list on your wiki

DrDaveD:
It may make more sense to add them to the standard OSG stratum 1s which are all set up for monitoring, etc.

PerilousApricot:
Sure, but I'm not OSG, so I can only offer the resources I have :)

DrDaveD:
But you do represent one of the biggest OSG customers (CMS) so probably all you need to do is ask.

rptaylor:
@PerilousApricot No problem. Our server meta info has a contact address but you can also use the support address here https://docs.alliancecan.ca/wiki/Technical_support  , but that is more for support requests.

Dave can certainly CC me on emails. 

We'll discuss internally but I think we'd be more inclined to have one of the OSG stratum servers mirror our repo(s) if that meets your needs and @DrDaveD agrees?
The main repo soft.computecanada.ca  is about ~ 1.5 TB.

DrDaveD:
As it turns out, the FNAL stratum 1 automatically replicates the OSG (at UNL) stratum 1 so we can't do only one stratum 1 unless it's BNL which does everything manually.  I recommend using both the OSG & FNAL stratum 1s.  All it requires is adding the stratum 0 URL into OSG topology associated with some VO, where the catchall is the osg VO.  The normal process for importing a repo into OSG includes asking BNL to manually add it but in this case since there are also stratum 1s in Canada I would recommend skipping that step.

DrDaveD:
I forgot to mention that the Sampa stratum 1 in South America also always automatically replicates everything on the OSG Stratum 1.

PR #271, opened 2024-11-26
Remove loading missing globus libraries from cvmfs for x.509 access to storage

DrDaveD:


etc/cvmfs/osgstorage-auth.conf
```diff
@@ -48,62 +48,13 @@ CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://xcachevirgo.pic.es:8443/"
...
-        if [ "$DISTROVERSION" = 6 ]; then
-            # choose the last OSG release to support el6
-            OSGRELEASE=3.4/current
-        fi
-        CVMFS_AUTHZ_OSG_WN_CLIENT="$OSGBASEDIR/$OSGRELEASE/el$DISTROVERSION-$DISTROARCH"
-    fi
+if [ ! -f /etc/grid-security/certificates/cilogon-basic.pem ]; then
+    # Get CA certs from the oasis repo
+    X509_CERT_DIR="$CVMFS_MOUNT_DIR/oasis.opensciencegrid.org/mis/certificates"
 else
```
jthiltges, line 55: Can we drop this workaround? Looks like it was [addressed](https://its.cern.ch/jira/browse/CVM-1083) in >= cvmfs 2.4.

etc/cvmfs/osgstorage-auth.conf
```diff
@@ -48,62 +48,13 @@ CVMFS_EXTERNAL_URL="$CVMFS_EXTERNAL_URL;https://xcachevirgo.pic.es:8443/"
...
     #  string, and cvmfs-2.3.2 tries to use the empty string as its CA dir
     X509_CERT_DIR=/etc/grid-security/certificates
 fi
-# The wrapper scripts here are mostly needed because LD_LIBRARY_PATH can't
-#  be exported.  See https://sft.its.cern.ch/jira/browse/CVM-1023
-# Other variables with a CVMFS_AUTHZ_ prefix will be passed to the helper
-#  with the prefix stripped off.
-CVMFS_AUTHZ_SEARCH_PATH="$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/libexec/authz"
-
+CVMFS_AUTHZ_SEARCH_PATH="$CVMFS_DIST/usr/libexec/cvmfs/authz"
```
jthiltges, line 60: Is setting this variable still necessary?

jthiltges (approved):
LGTM. Added some comments wondering about additional cleanup while we're working here.

etc/cvmfs/osgstorage-auth.conf
DrDaveD: It is.  I tried without it and it didn't work.

etc/cvmfs/osgstorage-auth.conf
DrDaveD: I thought about it but I'd rather leave it in.  I'm not 100% sure that nobody is still running that old of a client, although since the EL7 end of life there probably isn't anybody.  I'm also not sure how solid a default of /etc/grid-security/certificates is embedded.  It doesn't hurt to keep it in.

etc/cvmfs/osgstorage-auth.conf
DrDaveD: Ah but you're right, it does work if I also remove the `CVMFS_AUTHZ_SEARCH_PATH=` settings.  Removed.

etc/cvmfs/osgstorage-auth.conf
DrDaveD: Oh, and it's also needed because of the `X509_CERT_DIR=` settings in files that include this file, which is required in order for cvmfs to recognize the setting.  I updated the comment to reflect that.

djw8605:
Just a quick question, I see this commit removed the scitokens helper script.  But, I presume token auth still work, in some other manner?

DrDaveD:
> Just a quick question, I see this commit removed the scitokens helper script. But, I presume token auth still work, in some other manner?

Yes, it directly invokes /usr/libexec/cvmfs/authz/cvmfs_scitoken_helper so tokens still work.  x.509 still works to access the cvmfs metadata too if globus libraries are locally installed.

djw8605 (approved):
LGTM

jthiltges:
Published to config-osg.opensciencegrid.org as revision 142

PR #273, opened 2024-12-05
Remove loading missing globus libraries from cvmfs for x.509 access to storage (egi)

DrDaveD:
Cherry-pick #271 to the egi branch.

@alex-rg This will be for you to approve, merge, & deploy.  I recommend waiting until next week since it was just deployed on osg and we should get reports soon if it causes any problems for authenticated repositories.  

The affected authenticated repositories are ligo.osgstorage.org, ligo.storage.igwn.org, and virgo.storage.igwn.org.  These changes should not affect token access at all which is what they primarily use now.  In my testing the attempted loading of these globus libraries from oasis.opensciencegrid.org wasn't working anyway, it just caused disturbing messages in /var/log/messages, so I don't think this breaks anything new for x.509 cert or proxy access either.


alex-rg (approved):
LGTM.

PR #274, opened 2024-12-06
restore setup for RAL Stratum-1

jose-caballero:


DrDaveD:
It's close but not quite ready for this yet, I am discussing it with the admin team. Also it isn't good to make a change of this magnitude just before a weekend.

etc/cvmfs/domain.d/kek.jp.conf
```diff
@@ -5,9 +5,9 @@ CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
...
 else
-    CVMFS_SERVER_URL="http://cvmfs-stratum-one.cc.kek.jp:8000/cvmfs/@fqrn@;http://grid-cvmfs-one.desy.de:8000/cvmfs/@fqrn@;http://cvmfs-s1bnl.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-stratum-one.ihep.ac.cn:8000/cvmfs/@fqrn@;http://cvmfs-s1.hpc.swin.edu.au:8000/cvmfs/@fqrn@;http://sampacs01.if.usp.br:8000/cvmfs/@fqrn@"
+    CVMFS_SERVER_URL="http://cvmfs-stratum-one.cc.kek.jp:8000/cvmfs/@fqrn@;http://cernvmfs.gridpp.rl.ac.uk:8000/cvmfs/@fqrn@;http://grid-cvmfs-one.desy.de:8000/cvmfs/@fqrn@;http://cvmfs-s1bnl.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-s1fnal.opensciencegrid.org:8000/cvmfs/@fqrn@;http://cvmfs-stratum-one.ihep.ac.cn:8000/cvmfs/@fqrn@;http://cvmfs-s1.hpc.swin.edu.au:8000/cvmfs/@fqrn@;http://sampacs01.if.usp.br:8000/cvmfs/@fqrn@"
```
DrDaveD, line 10: While we're doing this let's change cernvmfs.gridpp.rl.ac.uk to cvmfs-egi.gridpp.rl.ac.uk for this domain, leaving the cernvmfs alias only for the cern.ch domain.

There's really little point to having the cernvmfs alias at all, but it is the name still used in the cvmfs-config-default rpm for the cern.ch domain so it should stay there.  The original idea of having two names was to be able to move the cern.ch repositories to a separate server if that was ever desired.  There's no point to having it be used for the kek.jp domain.

DrDaveD (changes requested):
This is the reverse of #264.  

After the one small change to this PR that I requested, I think now that the RAL stratum 1 service is in sufficiently good shape to be put into service early next week, even though there are more changes that should be done to it after that point.

jose-caballero:
I have changed the aliases to use cvmfs-egi only

DrDaveD (approved):

PR #278, opened 2025-02-24
Updating CVMFS config for the Pelican/OSDF sw

biozit:

PR #284, opened 2025-02-25
fixing port and removing hosts (egi)

biozit:
Import of #278 to the egi branch.

PR #285, opened 2025-02-25
fixing port and removing hosts (default)

biozit:
Import of #278 to the default branch.


DrDaveD (changes requested):
This is quite different than the egi branch in #284.  Please make all the CVMFS_EXTERNAL_URL settings the same as the egi branch.

biozit:
But this is the default branch.

DrDaveD:
Yes, and it is used in most regions of the world but not so much in the U.S., much like egi.

DrDaveD (approved):
Looks good to me now.  @vvolkl it's up to you to merge & deploy.  There is urgency because the current configuration fails at CERN, so please don't stage it in test mode for long.

vvolkl (approved):

PR #291, opened 2025-03-20
Fixing host name amst-fiona.nationalresearchplatform.org

biozit:


DrDaveD (approved):
I see that the new name is an alias for the old name, so functionally there is no difference at this time.  Is there a plan to change the value anytime soon?  If not we don't need to ask the admins to deploy it at this time, although it would still be good to get it also into the other 3 branches.

PR #295, opened 2025-03-28
update from fiona-r-uva.vlan7.uvalight.net to  amst-fiona.nationalres…earchplatform.org (egi)

biozit:
update from fiona-r-uva.vlan7.uvalight.net to  amst-fiona.nationalresearchplatform.org:


DrDaveD (approved):

DrDaveD:
@alex-rg There's no need to deploy this until there's another change because this only replaces a server name with its DNS alias.

PR #296, opened 2025-03-28
update from fiona-r-uva.vlan7.uvalight.net to  amst-fiona.nationalres…

biozit:
…earchplatform.org:

update from fiona-r-uva.vlan7.uvalight.net to  amst-fiona.nationalresearchplatform.org:


DrDaveD (approved):

DrDaveD:
@vvolkl There's no need to deploy this until there's another change because it only replaces a server name with its DNS alias.

PR #298, opened 2025-04-08
Use $RANDOM % 2 instead of $$ % 2 to avoid unbalance in pid namespaces

### LINKED ISSUE #297: Modify random number creation in default.conf to make it work inside docker
I'm testing the upcoming docker based app from https://lhcathomedev.cern.ch/lhcathome-dev/.
Recently I stumbled over an issue with cvmfs-config.cern.ch.

/cvmfs/cvmfs-config.cern.ch/etc/cvmfs/default.conf starts with this line:
`if [ "$(($$ % 2))" -eq 1 ]; then`

A common and elegant method to create pseudo random numbers based on the 
shell's own pid.
Here it is used for load balancing.

In most cases it can be assumed that the pid is either even or odd.
This does not apply to a shell started as docker entrypoint.
Here the pid is always '2'.
For CVMFS used from within a container this means the if-line above always 
evaluates to '0' and load balancing never happens.

To solve this default.conf in the cvmfs-config-default package should be modified like this:
`if [ "$((RANDOM % 2))" -eq 1 ]; then`

DrDaveD:
- Fixes #297

### DIFF
--- a/etc/cvmfs/default.conf
+++ b/etc/cvmfs/default.conf
@@ -1,4 +1,4 @@
-if [ "$(($$ % 2))" -eq 1 ]; then
+if [ "$((RANDOM % 2))" -eq 1 ]; then
   CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat"
 else
   CVMFS_PAC_URLS="http://grid-wpad/wpad.dat;http://wpad/wpad.dat;http://cernvm-wpad.cern.ch/wpad.dat;http://cernvm-wpad.fnal.gov/wpad.dat"


DrDaveD:
In my testing so far under cvmfs, the RANDOM in this expression is always returning 1.

computezrmle:
In a bash:
`for i in {1..20}; do echo "$((RANDOM % 2))"; done`
1
1
0
1
1
1
1
0
0
1
1
0
0
1
1
0
0
0
1
1

`for i in {1..20}; do echo "$(($RANDOM % 2))"; done`
0
0
0
0
1
1
1
0
0
1
0
1
0
0
0
1
0
1
1
1

DrDaveD:
Yes it was working for me interactively too, but I was only seeing it return 1 when I put it in a config repository's default.conf.

However, it looks like I was just getting unlucky and hadn't run enough trials, because I do now see the other answer there too.  Several times I ran 100 trials like this:
```
N=100
while [ $N -gt 0 ]; do
  let N-=1
  ls /cvmfs/atlas.cern.ch >/dev/null
  sudo cvmfs_talk -i atlas.cern.ch parameters|grep PAC|cut -d/ -f12
  sudo umount /cvmfs/atlas.cern.ch
done|sort|uniq -c
```
and I do see results on average close to 50/50.  So it's OK.

DrDaveD:
@vvolkl Please review

vvolkl (approved):

vvolkl:
Deployed in testing

vvolkl:
Deployed in production

PR #292, opened 2025-03-21
Fixing host name amst-fiona.nationalresearchplatform.org (osg)

biozit:
from fiona-r-uva.vlan7.uvalight.net to amst-fiona.nationalresearchplatform.org.  Cherry-pick of #291 to the osg branch.

PR #301, opened 2025-04-24
Use $RANDOM % 2 instead of $$ % 2 to avoid unbalance in pid namespaces (osg)

DrDaveD:
Cherry-pick #298 to the osg branch.


jthiltges:
Published to config-osg as revision 146, also including pending change https://github.com/cvmfs-contrib/config-repo/pull/292

PR #302, opened 2025-04-24
Use $RANDOM % 2 instead of $$ % 2 to avoid unbalance in pid namespaces (egi)

DrDaveD:
Cherry-pick #298 to the egi branch.


alex-rg (approved):
LGTM

PR #307, opened 2025-07-03
Add RAL Stratum-1 to the list of Stratum-1 servers for the EESSI repository

alex-rg:
RAL is becoming an official mirror for the repository.


DrDaveD (approved):
Note that this adds the configuration for all repositories in the domain, which is the right way to do it.  However I see that only software.eessi.io is replicated onto RAL even though there are also dev.eessi.io and riscv.eessi.io repositories on the other two stratum 1s.  That's not necessarily a problem, clients can detect that and failover to the next stratum 1, but replicating the others should probably be considered as well.

After this point replicating at least software.eessi.io to other WLCG stratum 1s should also be considered.

PR #315, opened 2025-08-19
When WPAD is enabled and CVMFS_USE_CDN is unset, set the latter to yes

DrDaveD:
See discussion in the [forum](https://cernvm-forum.cern.ch/t/cvmfs-unusable-from-cold-cache/546/10).

The osg & egi branches already do this and also if `CVMFS_HTTP_PROXY=DIRECT`, but @jblomer didn't want the DIRECT setting to go to the CDN by default.  If I recall correctly, that was the reason for the introduction of `CVMFS_CLIENT_PROFILE=single`.  However, the WPAD default response (when no squid is found) only works when the CDN is used.


vvolkl (approved):

vvolkl:
I was a bit too triggerhappy here, there's a typo (duplicate `fi`). Will fix in another PR.

PR #321, opened 2025-08-21
Fix typo in #315

vvolkl:


DrDaveD (approved):

DrDaveD:
Right.  In the osg branch there's two nested ifs and at the last minute I decided to collapse it into one layer but didn't complete removing the second layer.

vvolkl:
Deployed in testing

PR #330, opened 2025-09-09
add CDN servers for EESSI, use `eessi.science` CNAME for RAL Stratum 1 (master)

bedroge:
@DrDaveD Could you review this? If it looks fine, I'll open the same PR for the other relevant branches. Should I open one for `osg` too?

### DIFF
--- a/etc/cvmfs/domain.d/eessi.io.conf
+++ b/etc/cvmfs/domain.d/eessi.io.conf
@@ -1,10 +1,15 @@
 # These are here so cvmfs will notice them if common.conf sets them
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
 CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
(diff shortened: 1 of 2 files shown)


DrDaveD (approved):

DrDaveD (changes requested):
I didn't notice at first that the name of the non-CDN RAL server changed.  Please see the [ci check](https://github.com/cvmfs-contrib/config-repo/actions/runs/17581122156/job/49948380037?pr=330) and also make the corresponding change to geoipdb/server-cities.csv, adding the `,GB,Didcot,Oxfordshire` to the new line.

Yes after that you can make corresponding PRs for the default and egi branches.  The domain hasn't been approved yet for osg so that will need to wait.

DrDaveD (changes requested):
I'm sorry, but you also need to remove `cvmfs-ext.gridpp.rl.ac.uk` because the check makes sure that server-cities.csv has exactly the servers configured in the repo.

DrDaveD (approved):

PR #337, opened 2026-02-23
Add remaining U.S. OSDF caches (default)

DrDaveD:
The egi configuration repository configures only a subset of the Open Science Data Federation (OSDF) caches in the U.S., because people are expected to use the osg configuration there.  For the default configuration repository we have included the same subset as egi, under the assumption that people in the U.S. do not use cvmfs-config-default.  I have now found evidence that that assumption is wrong.  In particular, the Perlmutter HPC at NERSC has a good reason to use the default rpm, because it is a SUSE system and the OSG does not build SUSE rpms.  It's true that cvmfs-config rpms only include portable files, but still it's easy to see why system administrators would be reluctant to install elX rpms on a SUSE system.  There are very likely other installations in the U.S. that are also using cvmfs-config-default for various other reasons (including just not knowing which one they should use).

Instead, this PR increases the list of OSDF caches in osgstorage.org.conf to match the master branch which contained the superset of those listed by osg and egi.

### DIFF
--- a/etc/cvmfs/domain.d/osgstorage.org.conf
+++ b/etc/cvmfs/domain.d/osgstorage.org.conf
@@ -13,14 +13,27 @@ CVMFS_KEYS_DIR=$CVMFS_MOUNT_DIR/$CVMFS_CONFIG_REPOSITORY/etc/cvmfs/keys/osgstora
 CVMFS_USE_GEOAPI=yes
 CVMFS_SEND_INFO_HEADER=yes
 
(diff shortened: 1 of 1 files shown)


DrDaveD:
@vvolkl This is for you to merge and deploy.

vvolkl (approved):

vvolkl:
Deployed to testing, did one run of cloudtests.  Since that was fine, deployed to production just now.

PR #339, opened 2026-03-11
Fixing FQDNs (default)

biozit:

PR #342, opened 2026-03-13
Fixing hostname stashcache.t2.ucsd.edu -> ucsd-t2-cache.nationalresearchplatform.org (default)

biozit:


DrDaveD:
@biozit this replacement for #339 is still showing it is trying to re-merge old commits.

You may need to recreate your default branch from the upstream default branch, and then from there create a branch with just the one new commit.

PR #349, opened 2026-06-12
desy.de: preparing key rotation: adding upcoming key for clients

thdesy:
preparing for rotating the repository signing key. Multiple signing keys are accepted by clients and new commits are checked for all present public keys. RSA keys with 2048 bity sizes are the default, 4096 bit sized RSA keys worked in testing. Deploying both sizes with aim to go for 4096b and fall back to 2048b in case of problems.


DrDaveD (approved):
Looks fine.  Please also make cherry-pick PRs for the osg, egi, and default branches.

thdesy:
thanks - I have just added cherry-picks for the three branches

thdesy:
hmmm, I just noticed that our current (still active) public key is not in the default branch. Meaning that only the upcoming would end up in it, or? Should I create a new PR to for the current `desy.de.pub` key?

While checking, I noticed that in the OSG config there is also in the `desy.de` key dir the `opensciencegrid.org` repo key
 `/cvmfs/config-osg.opensciencegrid.org/etc/cvmfs/keys/desy.de/opensciencegrid.org.pub`
 should probably do not harm, but might be confusing (maybe in a combined MR or in two separate?)

DrDaveD:
There's no config for desy.de in the default branch at all, nothing in domain.d either.  Apparently nobody has requested it before that uses the cvmfs-config-default distribution.  You would need to make a separate PR for that and see if Valentin agrees with it (I don't see why he wouldn't).

The reason for the OSG key in the osg branch is historical.  In the early years of cvmfs on OSG there were some influential people who wanted the ability for OSG central administrators to be able to publish emergency updates in every repository that they distributed to their users in case of a security breach (other than cern.ch repos, which was grandfathered).  Those people left and the policy was rescinded quite a few years ago.  But the OSG keys haven't all been cleaned out.

PR #359, opened 2026-08-13
Add ihep.ac.cn domain, keys and config

CatalinCondurache:
Added public key and config file for ihep.ac.cn domain.


etc/cvmfs/domain.d/ihep.ac.cn.conf
```diff
@@ -0,0 +1,13 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1ral-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1nikhef-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1ihep-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1swinburne-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1sampa-cvmfs.openhtc.io:8080/cvmfs/@fqrn@"
+else
+    CVMFS_SERVER_URL="http://cvmfs-stratum-one.ihep.ac.cn:8000/cvmfs/@fqrn@;http://cvmfs-egi.gridpp.rl.ac.uk:8000/cvmfs/@fqrn@"
```
DrDaveD, line 10: Is this correct that these repos are only replicated at IHEP and RAL?  Do we need to recruit more stratum 1s to replicate them?  Where are the users of these repositories?  Usually we try to get heavily used repos replicated on 2 stratum ones in each continent where they're used.

I see that RAL has only replicated 5 of the ihep.ac.cn repositories, and there are a lot more on the IHEP stratum one.  That may be OK if those are the only ones that are used in Europe.  I verified that I was able to mount all 5 of those repositories from RAL using the key you included.

etc/cvmfs/domain.d/ihep.ac.cn.conf
```diff
@@ -0,0 +1,13 @@
+# These are here so cvmfs will notice them if common.conf sets them
+CVMFS_USE_CDN="$CVMFS_USE_CDN"
+CVMFS_HTTP_PROXY="$CVMFS_HTTP_PROXY"
+CVMFS_FALLBACK_PROXY="$CVMFS_FALLBACK_PROXY"
+. ../common.conf
+
+if [ "$CVMFS_USE_CDN" = "yes" ]; then
+    CVMFS_SERVER_URL="http://s1ral-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1nikhef-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1bnl-cvmfs.openhtc.io/cvmfs/@fqrn@;http://s1fnal-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1ihep-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1swinburne-cvmfs.openhtc.io:8080/cvmfs/@fqrn@;http://s1sampa-cvmfs.openhtc.io:8080/cvmfs/@fqrn@"
```
DrDaveD: This list should only contain the openhtc.io equivalents of the stratum ones where the repos are replicated, that is, the ones in the `else` case here.  Please shorten this list to just ihep and ral for now.

DrDaveD (changes requested):

etc/cvmfs/domain.d/ihep.ac.cn.conf
CatalinCondurache: AFAIK there are two JUNO groups (Catania INFN and RAL STFC) looking for a better access to ihep.ac.cn repos. 
By better access I mean a default CVMFS client configuration in place that would allow automated mount of these repos.
And indeed there are only 5 repos mirrored by RAL, not sure we need NIKHEF as a second mirror for them for the time being.

etc/cvmfs/domain.d/ihep.ac.cn.conf
DrDaveD: @CatalinCondurache this is waiting on you.

CatalinCondurache: Updated as requested, thank you!

DrDaveD (approved):
@alex-rg This is for you to merge and deploy.

@CatalinCondurache Please also make a PR for the master branch that does the same changes.  Usually we do that with git cherry-pick.

